Issue #24, opened 2026-07-11
Consider Durable Objects for per-workspace usage enforcement


zachdunn:
## Context

We now have an **observe-first** workspace usage ledger in D1 (`workspace_usage`), updated best-effort from the shared put/delete path and exposed via `GET /v1/:workspace/usage` (#23). Tracking issue for quotas/retention foundations: #22.

That design is the right first step for measurement. When we move from **observing** counters to **enforcing** hard budgets (or running per-workspace retention jobs), Cloudflare **Durable Objects** are a strong candidate for the ledger authority.

This issue captures the DO option so we can revisit it in a follow-up without blocking current work.

## Why DOs fit workspace ledgers

A workspace is already a single tenant boundary (one prefix, shared by many tokens). A DO per workspace maps cleanly:

| Concern | D1 (current) | Durable Object per workspace |
| --- | --- | --- |
| Concurrent put/delete metering | Racey / last-writer-wins at high parallelism | Serialized execution — increments correct by construction |
| Enforce “may I store +N bytes?” | Read → decide → write (TOCTOU) | `tryReserve` / `release` in one place |
| Period reset, retention | Cron or on-read fudge | Alarms on that workspace’s DO |
| Multi-token attribution | Optional side table | Same DO; optional event log (e.g. DO SQLite) |
| Fail-closed under budget | Awkward with best-effort D1 | Natural: storage path awaits DO RPC |

Sketch:

```text
idFromName(workspaceName) → WorkspaceUsage DO

PUT  → head (overwrite delta) → do.tryReserve({ bytesDelta, uploads }) → R2 → on failure do.release
GET /usage → do.snapshot()
```

Net stored bytes still ultimately live in R2; DOs fix **counter concurrency and enforcement**, not “object deleted outside the API.” Occasional `listAll` reconcile under the workspace prefix remains a backstop either way.

## When this becomes worth it

Prefer DO when we start any of:

- [ ] Hard `maxStorageBytes` / monthly upload budget enforcement under concurrent agents
- [ ] Reserve-then-upload (reject before accepting a large body into isolate memory)
- [ ] Per-workspace retention/TTL driven by alarms (free-tier expiry, idle cleanup)
- [ ] Strong single “source of write budget” without inventing locks on D1

Stay on D1 while:

- Observe-first is enough and QPS stays sparse (PR screenshots / agent assets)
- Drift fixed by reconcile is acceptable
- We want minimal moving parts (auth is already D1)

[...]

Issue #22, opened 2026-07-11
Workspace storage quotas, upload budgets, and retention foundations


zachdunn:
## Why

Hosted multi-tenant use needs more than per-object size and write rate limits. Without **storage accounting**, **upload budgets**, and optional **retention**, a shared bucket can grow without bound even when individual uploads stay under the 25 MiB cap. This is foundational multi-tenant safety and operator DX — useful whether a workspace is internal, OSS, or a future hosted tier.

Today we already have (good starting points):

- `maxUploadBytes` / content-type allowlist (`apps/api/src/guards.ts`)
- Per-workspace write rate limiting (`WRITE_LIMITER`)
- Workspace records in KV (`WorkspaceRecord` in `apps/api/src/workspace.ts`)

What’s missing is cumulative usage, configurable budgets, lifecycle, and clear client errors when a workspace is over budget.

## Goals

1. **Account for stored usage** per workspace (at least total bytes; ideally object count).
2. **Enforce configurable budgets** on the workspace record (storage bytes and/or uploads per period), merged the same way as `maxUploadBytes`.
3. **Retention / lifecycle hooks** so prefixes or workspaces can expire objects (e.g. TTL after upload, or expire-after-idle) without hand-deleting forever.
4. **Agent-friendly failures** — structured API errors and CLI/MCP messages that say what limit was hit and what to do (delete objects, raise the budget, wait for period reset).
5. **Operator visibility** — a way to inspect usage for a workspace (admin and/or authenticated client).

Non-goals for this issue:

- Public marketing or commercial packaging
- Billing integration / payment providers
- Transforming media or bandwidth-based product limits (R2 egress is free; Class B is an internal cost, not a first customer-facing meter)

### Workspace record fields (sketch)

Extend `WorkspaceRecord` with optional overrides, all optional so existing workspaces stay unlimited until set:

- `maxStorageBytes?: number` — hard cap on total stored bytes under the workspace prefix
- `maxUploadsPerPeriod?: number` + period (e.g. calendar month or rolling 30d)
- `retentionDays?: number` — optional default TTL for new objects (or a separate policy object later)
- Existing: `maxUploadBytes`, `allowedContentTypes`

### Storage accounting

Options to evaluate:

- **Increment/decrement counters** on put/delete (KV or DO) — fast enforcement, need care on failures/races and list-import drift
- **Periodic reconcile** from `list` / bucket inventory under the workspace prefix
- Hybrid: counters for enforcement + daily reconcile job

[...]

PR #26, opened 2026-07-11
feat(api): reconcile usage + retention purge

zachdunn:
## Summary

High-value follow-ups before Durable Objects (#24):

1. **Reconcile** — `POST /v1/:workspace/usage/reconcile` (`files:write`) walks storage via files-sdk `listAll`, replaces ledger `bytes`/`objects`, **keeps** `uploadsInPeriod`.
2. **Retention** — optional `retentionDays` on the workspace KV record; `POST /v1/:workspace/usage/purge-expired` (`files:delete`) deletes objects older than that window (by last-modified), then reconciles. Skips cleanly when unset.
3. **Config** — `pnpm workspace:limits <ws> --retention-days 90` / `--clear-retention-days`.

No automatic cron yet — ops/CI can call purge when desired.

Relates to #22. Deliberately not #24 (DO).

## Test plan

- [x] API unit/route tests (drift fix, purge old/keep new, skip without retention)
- [x] Full API + MCP suites green
- [ ] After merge: reconcile a prod workspace; optionally set retention and dry-run logic via purge on a test prefix

### DIFF
--- a/apps/api/scripts/set-workspace-limits.mjs
+++ b/apps/api/scripts/set-workspace-limits.mjs
@@ -8,7 +8,9 @@
  *     [--max-storage 25GB] \
  *     [--max-uploads-per-month 10000] \
  *     [--max-upload-bytes 25MB] \
+ *     [--retention-days 90] \
  *     [--clear-max-storage] [--clear-max-uploads-per-month] [--clear-max-upload-bytes] \
+ *     [--clear-retention-days] \
  *     [--local]
  *
  * Units: bare number = bytes (or count for uploads). Also accepts KB/MB/GB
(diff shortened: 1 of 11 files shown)


zachdunn:
### files-sdk alignment (follow-up commit)

| Surface | Use? | Why |
| --- | --- | --- |
| \`listAll()\` | **Yes** | Reconcile + purge walk; prefix already on workspace \`Files\` instance |
| \`delete(keys[])\` bulk | **Yes** | Purge batches of 500 — native multi-delete on R2/S3 |
| \`usage()\` plugin | No | In-memory per isolate; not durable net storage |
| \`softDelete()\` | No | Recycle bin / trash prefix, not TTL retention |
| Lifecycle via \`files.raw\` | No (for now) | Bucket-wide; shared-bucket multi-tenant needs per-workspace days |

PR #40, opened 2026-07-11
feat: apply default limits when creating a workspace

zachdunn:
## In plain terms

When you create a new workspace, it should **not** start with unlimited storage and unrestricted paths. New workspaces now get the same **shared/agent** limits we already use on production `default` (25 GB, 10k uploads/month, size caps, 90-day retention, and the `f` / `screenshots` / `gh` key roots).

### What it does

- **`pnpm workspace:add`** applies a default limit template automatically
- You can still **override** individual fields with flags, or start unlimited with **`--no-default-limits`**
- Existing workspaces are **unchanged** (use `workspace:limits` for those)

### What it is not

- Not a runtime change on the Worker (create path is the operator script only)
- Not a new npm package release
- Not forcing throwaway-tier limits — this is the shared/agent profile

### How to try it

```bash
pnpm workspace:add demo-ws --local
# prints limits JSON including storage, retention, allowed prefixes

pnpm workspace:add free-ws --no-default-limits --local
# unlimited (legacy behavior)
```

### Technical notes

- Template: `apps/api/scripts/workspace-limit-defaults.json` (+ thin `.mjs` loader)
- `add-workspace.mjs` merges template first; explicit flags override; `unlimited` clears a field
- Also supports `--retention-days` on create (was missing before)
- Docs: `docs/ops.md`, `docs/workspaces.md`, `AGENTS.md`, `apps/api/README.md`

### Test plan

- [x] `pnpm --filter @uploads/api test`
- [x] `pnpm check`
- [x] Smoke-import `sharedAgentLimitFields()`
- [ ] Optional: create a local workspace and confirm printed limits

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -36,7 +36,7 @@ pnpm dev:web             # Astro site
 pnpm typecheck           # wrangler types + tsc across workspaces
 pnpm run deploy          # all workers; or deploy:api / deploy:web / deploy:mcp
 pnpm workspace:add <name> [--bucket <bucket>] [--binding X] [--local] \
-  [--max-storage 25GB] [--max-uploads-per-month N] [--max-upload-bytes 25MB]
+  [--no-default-limits] [--max-storage …]   # shared/agent limit template by default
 pnpm workspace:limits <name> [--max-storage …] [--max-video-bytes …] \
   [--allowed-prefixes default|f,screenshots,gh] [--max-key-depth 8] \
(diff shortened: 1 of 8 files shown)


zachdunn:
Dropped **retention** from the default create template (and cleared `retentionDays` on production `default`). Caps + key policy remain; expiry is opt-in via `--retention-days` for throwaway workspaces.

Issue #66, opened 2026-07-11
First-class media galleries linked to external work


zachdunn:
## In plain terms

Make media galleries a first-class uploads.sh resource: a stable collection created once, populated by humans or agents, and optionally linked to one or more GitHub PRs/issues or future external work items.

The gallery owns an opaque uploads.sh identity such as `uploads.sh/g/<id>`. GitHub coordinates are many-to-many external references and lookup indexes, not gallery coordinates. This lets galleries exist before a PR, span multiple PRs, survive repository changes, and later integrate with systems such as Linear without reshaping the storage model.

## Launch outcome

A user or agent can:

1. Create a gallery and receive a stable public URL.
2. Upload or add ordered media to it.
3. Link it to one or more GitHub items without installing an App.
4. Find every gallery related to a GitHub coordinate.
5. Share or maintain a managed GitHub comment containing gallery links.

## Product constraints

- Galleries are public in v1; private GitHub repositories do not imply private media.
- Object bytes remain in provider-backed storage through `createStorage()`.
- Gallery membership references object keys and does not duplicate bytes.
- Relationships are many-to-many and provider-neutral.
- Existing coordinate-keyed `gh/...` attachments remain compatible.
- GitHub custom autolinks are an optional advanced integration, not a launch dependency; availability and administration requirements make them a specialized feature.

## Work breakdown

- [x] #56 — Gallery contract: define identity, ownership, items, and visibility
- [x] #61 — Gallery persistence: add workspace-scoped galleries, items, and references
- [x] #59 — Gallery API: authenticated CRUD and ordered media operations
- [x] #62 — External references: many-to-many provider-neutral gallery links
- [ ] #60 — Public gallery: render a stable read-only media collection
- [ ] #63 — CLI: create and manage first-class galleries
- [x] #64 — MCP: expose gallery operations and reference lookup
- [x] #58 — GitHub linking: associate and find galleries by PR or issue
- [ ] #65 — GitHub comments: show linked galleries with direct attachments
- [ ] #57 — Gallery launch hardening: lifecycle, retention, limits, and contract tests

## Dependency shape

The contract informs persistence. Persistence enables gallery and reference APIs. Those APIs enable the web, CLI, and MCP experiences. GitHub linking builds on general references; managed comments build on linking and public gallery URLs. Lifecycle and end-to-end work close the launch milestone.

## Explicitly later

- GitHub custom-autolink setup
- GitHub App and automatic backlink detection
- GitHub-authenticated private galleries
[...]

Issue #62, opened 2026-07-11
External references: many-to-many provider-neutral gallery links


zachdunn:
## In plain terms

Add provider-neutral external references so a gallery can be associated with GitHub work today and other systems later without changing gallery identity.

## Initial model

For GitHub, accept coordinates such as `buildinternet/uploads#123` and normalize them into provider, repository scope, item type, and identifier fields. GitHub PR and issue numbers share a sequence, so callers should not need to know which kind exists merely to link or look it up.

## Acceptance criteria

- [ ] Add and remove references from a gallery
- [ ] Find every gallery linked to a normalized external reference
- [ ] Many-to-many relationships are supported
- [ ] Repeating the same link operation is idempotent
- [ ] Provider parsing is separate from the stored relationship model
- [ ] Reference values are validated and length-bounded
- [ ] Tests cover tenant isolation, duplicates, normalization, and multiple matches
- [ ] The model leaves room for Linear/Jira without implementing them

## Depends on

Gallery contract and persistence.

Issue #61, opened 2026-07-11
Gallery persistence: add workspace-scoped galleries, items, and references


zachdunn:
## In plain terms

Add the durable, workspace-scoped data model for galleries, ordered gallery items, and external references.

## Proposed shape

- `galleries`: opaque ID, workspace, title/description, timestamps, visibility state
- `gallery_items`: gallery, stored object key, stable item ID, explicit position
- `external_references`: gallery plus normalized provider/scope/type/identifier fields

Object bytes remain in provider-backed storage. Gallery membership references keys and must not copy media.

## Acceptance criteria

- [ ] D1 migrations add the schema and useful lookup indexes
- [ ] Every query and constraint preserves workspace isolation
- [ ] Gallery IDs are opaque and URL-safe
- [ ] Items support deterministic ordering and defined duplicate behavior
- [ ] External-reference duplicates are prevented or idempotent
- [ ] Deleting a gallery does not silently delete stored objects
- [ ] Local migration and focused persistence tests pass
- [ ] Operator migration documentation is updated if needed

## Depends on

The gallery product/API contract.

Issue #60, opened 2026-07-11
Public gallery: render a stable read-only media collection


zachdunn:
## In plain terms

Build the public, read-only gallery page at a stable opaque URL such as `https://uploads.sh/g/<gallery-id>`.

## Experience

Show the gallery title and description, ordered media previews, filenames, and original-file links. Handle images, video, ordinary files, empty galleries, and media removed by retention without breaking the gallery URL.

## Acceptance criteria

- [ ] Responsive gallery renders from the canonical opaque ID
- [ ] Missing, expired, and deleted media have intentional states
- [ ] No workspace bearer token or arbitrary listing capability reaches the browser
- [ ] Rendering is safe for untrusted titles, filenames, and metadata
- [ ] Restrictive CSP and `noindex` are applied for the public-only v1 posture
- [ ] SVG remains unsupported in accordance with upload policy
- [ ] Accessibility and social/share metadata receive a basic pass
- [ ] Rendered pages are visually checked at desktop and mobile widths

## Depends on

Gallery API and persistence.

zachdunn:
Shipped in #71, with item pages in #88. Public `/g/<id>` (and `/g/<id>/<item>`) render a stable read-only collection with CSP/noindex and missing-media states. Closing this child; leftover hardening stays on #57.

PR #71, opened 2026-07-11
feat: add first-class media galleries

zachdunn:
## In plain terms

This introduces first-class public media galleries that can be linked to GitHub work without making a gallery's identity or lifetime depend on a PR or issue. A gallery gets its own stable URL, and GitHub references are optional relationships that can be added later or shared across multiple galleries.

## What it does

- Defines the gallery contract, lifecycle, privacy boundary, and URL model.
- Adds D1 persistence with ordered items, optimistic concurrency, soft deletion, and tenant isolation.
- Adds authenticated gallery management plus a deliberately narrow public read endpoint.
- Normalizes GitHub issue and PR references and supports paginated reverse lookup.
- Adds an on-demand `/g/<gallery-id>` web experience for images, videos, ordinary files, missing media, and empty galleries.
- Applies no-index, no-referrer, CSP, frame-blocking, and `no-store` controls to public gallery responses.

This does not make GitHub coordinates the gallery's canonical identity, expose external references publicly, or require GitHub custom autolinks. Autolinks can remain a specialized enhancement for eligible GitHub plans.

## Technical notes

- Galleries and references are workspace-scoped in D1.
- Public DTOs exclude workspace names, object keys, storage provenance, hashes, and external references.
- The web app now uses Astro's Cloudflare adapter while existing pages remain prerendered.
- Cloudflare Images is not required; gallery media is served from its existing public storage URL.

## Test plan

- [x] API suite: 143 tests
- [x] Public gallery web tests: 5 tests
- [x] Repository lint and format checks
- [x] Repository-wide TypeScript checks
- [x] Astro production build
- [x] Wrangler deployment dry-run
- [x] Local HTTP status, cache, and security-header checks
- [x] Independent correctness/privacy/security review
- [ ] Inspect rendered gallery states at desktop and mobile widths; browser automation could not attach to the WSL-mounted local preview

Tracks #66. Implements the foundation described by #56, #61, #59, #62, and #60.

## Summary by CodeRabbit

- **New Features**
  - Added gallery creation, editing, deletion, item management, ordering, pagination, and external-reference linking.
  - Added public gallery pages at `/g/<id>` with responsive media previews and safe fallback states.
  - Added unauthenticated public gallery API access with privacy and security protections.
  - Added GitHub reference parsing and reverse gallery lookup.

- **Documentation**
  - Documented gallery endpoints, visibility, versioning, limits, ordering, and reference behavior.

- **Chores**
  - Updated the minimum supported Node.js version to 24.


zachdunn:
Pulled the branch down and tested it end to end locally (local D1 + KV + R2 via `wrangler dev`, seeded workspace, real uploads, then the built web worker pointed at the local API). All 143 API tests and 5 web tests pass, and the API behaves as documented: create/patch/delete with optimistic concurrency, idempotent duplicate adds, `gallery_object_not_found` for unknown keys, case-normalized `by-reference` lookup, soft-deleted galleries vanish from the public endpoint, and the tenant boundary holds (`401` cross-workspace, public DTO omits workspace/keys/references). The `/g/<id>` page renders the populated, empty, not-found (404), and API-down (503) states with all the promised security headers, and desktop + mobile layouts look good — screenshots below cover the test-plan item that was left unchecked.

Two issues found, one of them blocking:

**1. Blocking: real browser visits to `/g/<id>` get the static 404 page.** Workers assets treat requests with `Sec-Fetch-Mode: navigate` specially: with `compatibility_date` ≥ 2025-04-01 (`assets_navigation_prefers_asset_serving`), a navigation that misses an asset is answered by `not_found_handling` instead of invoking the worker. This PR removed `run_worker_first` from `apps/web/wrangler.jsonc`, so every browser navigation to a gallery — the entire point of the feature — 404s, while curl/tests pass because they don't send `Sec-Fetch-Mode`. That's why the local HTTP checks in the test plan all passed. The regression also hits `/.well-known/agent-skills/index.json` for browser navigations. Verified locally:

```
plain curl:            GET /g/gal_… → 200
Sec-Fetch-Mode: navigate → 404 (wrangler logs: "`Sec-Fetch-Mode: navigate` header present - using `not_found_handling` behavior")
```

Fix (verified locally — after rebuild both paths return 200 for navigations, static pages unaffected):

```jsonc
"assets": {
  "binding": "ASSETS",
  "directory": "./dist",
  "not_found_handling": "404-page",
  "run_worker_first": ["/g/*", "/.well-known/agent-skills/index.json"],
},
```

**2. Blocking-ish: the build no longer runs on the repo's pinned Node.** `.nvmrc` pins 22.14.0, but the `@astrojs/cloudflare` adapter chain pulls in vite 8, which uses `node:module`'s `registerHooks` (added in Node 22.15). `astro build` fails on 22.14.0 with `The requested module 'node:module' does not provide an export named 'registerHooks'`; it builds fine on 24.x. Bump `.nvmrc` (and confirm what Workers Builds CI resolves) as part of this PR.

Non-blocking observations, take or leave:

- `GET /public/galleries/:id` has no rate limit and no `Cache-Control`, and each request does up to 2×N storage calls (`exists` + `head` per item, N ≤ 100) plus a D1 read. A gallery linked from a busy PR could generate a lot of R2 subrequests. Cheap wins: collapse `exists`+`head` into one `head` with a catch, and/or put a short `max-age` on the public JSON (same reasoning as the `max-age=60` upload-freshness change in #10).
- `hydrateGalleryItems` throws a 503 for the *whole* gallery if any one object exists but `publicUrl()` returns null (e.g. a workspace's `publicBaseUrl` is removed). Degrading that item to `missing` would keep the rest of the gallery up.
- The PR body says public DTOs exclude workspace names, and the fields do — but for shared-bucket workspaces the item `url` embeds the workspace prefix (`…/testws/shots/test.png`). That's inherent to the existing public storage URLs, just worth a line in the spec so nobody assumes the workspace is anonymous.
- Every write handler re-fetches and re-hydrates the full gallery (`ownerGallery`) for its response — another full head-sweep per mutation. Fine at N=100, just noting the pattern before it gets copied somewhere hotter.

The concurrency design itself held up well under poking — the batch INSERT…SELECT guards (version + dedupe + limit in SQL) classify every failure mode I could trigger into the right error code.

Rendered gallery states (local, with the `run_worker_first` fix applied; the broken thumbnail is expected — the seeded test object points at a fabricated public domain):

| Desktop | Mobile |
|---|---|
| Public gallery page, desktop | Public gallery page, mobile |

zachdunn:
Re-tested the revisions (e7c313e, b65440f) locally: `astro build` now succeeds on the pinned Node 24.18.0, and with `run_worker_first` in place a real browser navigation (`Sec-Fetch-Mode: navigate`) returns 200 for `/g/<id>` and `/.well-known/agent-skills/index.json`, 404 for unknown galleries and non-existent paths, with static pages unaffected. Full suites still pass (143 API + 5 web). Both blocking findings are resolved.

PR #81, opened 2026-07-12
fix: bound remaining local wrangler CLIs against orphan hangs

zachdunn:
## In plain terms

Local `wrangler … --local` boots miniflare. When an agent or shell times out and only kills the wrapper, the Node child can reparent and hang — we have seen multi‑GB RSS from a stuck `kv key get`. Bootstrap/doctor already avoid unbounded existence checks; this PR bounds the **remaining scripted** paths that still shell out to wrangler.

## What it does

- Adds `apps/api/scripts/run-timed.mjs` (GNU `timeout`/`gtimeout` preferred, Node spawn timeout fallback).
- Routes `workspace:add` / `workspace:limits` KV put/get through timed helpers (local ~30–60s, remote 120s).
- Caps `migrate:d1:local` at 60s via the same runner.
- Documents the operator path in AGENTS.md and docs/ops.md.

## What it is not

- Not a Cloudflare config change — Wrangler has no built-in wall-clock flag for one-shot `kv`/`d1` CLIs (skills + docs confirm this is process hygiene, not a wrangler.jsonc knob).
- Does not replace intentional long-lived `wrangler dev`.
- Does not change remote API behavior beyond a generous 120s ceiling on scripted KV ops.

## How to try it

```bash
# should complete quickly under the 60s cap
pnpm --filter @uploads/api run migrate:d1:local

# read-only limits dump via timed local get
pnpm workspace:limits default --local

# timeout path (expect exit 124)
node apps/api/scripts/run-timed.mjs 1 -- sleep 5
```

## Technical notes

- Prefer process-group kill via `timeout --kill-after=5s` so miniflare grandchildren die with the wrapper.
- Doctor/bootstrap still use SQLite-first registry checks (`scripts/lib-local.sh`) for existence.
- Sibling tracking: [releases#2121](https://github.com/buildinternet/releases/issues/2121), [room-configurator#333](https://github.com/buildinternet/room-configurator/issues/333), [either#794](https://github.com/buildinternet/either/issues/794).

## Test plan

- [x] `run-timed.mjs 5 -- echo ok` → 0
- [x] `run-timed.mjs 1 -- sleep 5` → 124
- [x] `workspace:limits default --local` succeeds
- [x] `migrate:d1:local` succeeds under wrapper
- [x] no leftover `wrangler.*(kv|d1).*--local` after runs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of local workspace management and D1 migrations by enforcing time-bounded execution to prevent hung local processes.
  * Added clearer, more specific messaging when local operations time out or fail to persist changes.
  * Improved robustness of local KV read/write operations used during workspace setup and limits updates.
* **Documentation**
  * Expanded local operations runbook guidance with recommended timed/guarded workflows for workspace limits and database migrations.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -79,6 +79,9 @@ path. Prefer:
 - **`pnpm doctor` / `pnpm bootstrap`** for “is local `default` registered?” —
   they check miniflare’s on-disk SQLite first and only fall back to a
   **time-bounded** wrangler call (`scripts/lib-local.sh`).
+- **`pnpm workspace:add` / `workspace:limits` / `migrate:d1:local`** go through
+  `apps/api/scripts/run-timed.mjs` (local get/put ~30–60s, D1 migrate 60s).
+  Do not reimplement bare `execFileSync(wrangler … --local)`.
 - When you must run wrangler yourself, wrap it:  
   `timeout 20s pnpm --filter @uploads/api exec wrangler kv key get ws:default --binding REGISTRY --local`
 - Never leave bare `wrangler kv|d1 … --local` running in the background. If an
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -82,8 +82,8 @@ path. Prefer:
 - **`pnpm workspace:add` / `workspace:limits` / `migrate:d1:local`** go through
   `apps/api/scripts/run-timed.mjs` (local get/put ~30–60s, D1 migrate 60s).
   Do not reimplement bare `execFileSync(wrangler … --local)`.
-- When you must run wrangler yourself, wrap it:  
-  `timeout 20s pnpm --filter @uploads/api exec wrangler kv key get ws:default --binding REGISTRY --local`
+- When you must run wrangler yourself, wrap it in the timed runner:  
+  `node apps/api/scripts/run-timed.mjs 20 -- pnpm --filter @uploads/api exec wrangler kv key get ws:default --binding REGISTRY --local`
 - Never leave bare `wrangler kv|d1 … --local` running in the background. If an
   agent times out a command, **kill the process group** (`pkill -f 'wrangler.*--local'`
   only after confirming PIDs), not just the shell wrapper.
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -20,7 +20,7 @@
   "scripts": {
     "dev": "wrangler dev",
     "migrate:d1": "node --env-file-if-exists=../../.env node_modules/wrangler/bin/wrangler.js d1 migrations apply DB --remote",
-    "migrate:d1:local": "CI=1 wrangler d1 migrations apply DB --local",
+    "migrate:d1:local": "CI=1 node scripts/run-timed.mjs 60 -- wrangler d1 migrations apply DB --local",
     "deploy": "pnpm run migrate:d1 && node --env-file-if-exists=../../.env node_modules/wrangler/bin/wrangler.js deploy",
     "types": "wrangler types",
     "workspace:add": "node --env-file-if-exists=../../.env scripts/add-workspace.mjs",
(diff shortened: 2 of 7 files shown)


zachdunn:
Reviewed — the approach is right. The one substantive gap found in review (Node fallback) is now **fixed in 5ed22ac**, along with a wrangler bump to ^4.110.0. Leaving the full write-up here since this PR is the reference for applying the same mitigation in other repos (releases#2121, room-configurator#333, either#794).

## Why the approach is sound

The failure mode (agent/shell timeout kills the wrapper; `pnpm → wrangler → workerd/miniflare` grandchildren reparent and sit at multi-GB RSS) has no wrangler config knob — one-shot `kv`/`d1` commands have no wall-clock flag, so a process-hygiene wrapper is the correct layer. Preferring `timeout --kill-after` is the right primary mechanism: `timeout` puts the child in its own process group and signals the whole group, so miniflare grandchildren die with the wrapper. BSD `timeout` (stock macOS 13+) accepts `--kill-after=5s` and exits 124, so the mixed GNU/BSD assumption holds.

## Fixed in 5ed22ac: fallback now kills the process group

As originally written, when no `timeout` binary was found, `spawnSync(..., { timeout, killSignal: "SIGKILL" })` SIGKILLed **only the direct child — `pnpm`**. SIGKILL doesn't propagate, so `wrangler`/`workerd` were instantly orphaned — the fallback *created* the exact zombie this PR exists to prevent.

Now: the sync library fallback re-invokes the script's own CLI, which does a **detached spawn** (child leads its own process group) and on deadline sends `SIGTERM` to `-pgid`, then `SIGKILL` after a 5s grace. Parent SIGINT/SIGTERM are forwarded to the group. Also added a `win32` guard (System32's `timeout.exe` is a wait command, not a killer) and a `RUN_TIMED_FORCE_FALLBACK=1` escape hatch for testing the fallback path. Verified: `RUN_TIMED_FORCE_FALLBACK=1 run-timed.mjs 2 -- sh -c 'sleep 60 & sleep 60'` exits 124 with no surviving grandchildren, via both the CLI and the library entry points.

## Remaining nits (non-blocking)

- In `set-workspace-limits.mjs`, a timeout on the **put** throws uncaught (only the get is inside the try/catch) — stack trace instead of the nice `fail(...)`. Cosmetic.

## Playbook for other repos: keeping local wrangler memory in check

There is still no first-class memory-limit knob for local workerd, so process hygiene is the right layer everywhere. Beyond porting `run-timed.mjs`, in rough priority order:

1. **RSS watchdog / OS memory cap as a backstop.** Wall-clock timeouts miss the *runaway-while-alive* case (actively leaking but under the deadline). Linux: `systemd-run --user --scope -p MemoryMax=2G …`. macOS has no clean cgroup equivalent; `run-timed.mjs` could grow an optional watchdog that polls `ps -o rss= -g <pgid>` and group-kills past a threshold.
2. **Dev-registry hygiene.** Every `wrangler dev` writes to `~/.wrangler/registry/` and watches it; stale entries from killed sessions cause other dev processes to retry proxy connections. A doctor step should clear registry files whose debug ports no longer answer.
3. **One dev server, not many.** The biggest real-world source of multi-GB usage is accumulation: several `wrangler dev` instances from successive agent sessions, each holding its own workerd + inspector state. Repo guidance should say "check for an existing listener (`lsof -ti :<port>`) before starting another."
4. **Watch-scope bloat.** `wrangler dev` rebuild loops leak memory if custom `persist-to` or state paths end up inside watched source dirs. Defaults are fine; audit custom paths.
5. **Keep wrangler current.** workerd/miniflare leak fixes land release-to-release (DO hibernation, inspector sessions). This PR bumps to ^4.110.0; other repos should pin-and-bump deliberately too.
6. **Prefer in-process miniflare for tests** (`@cloudflare/vitest-pool-workers` or the `Miniflare` API with explicit `await mf.dispose()`) over shelling out to the CLI — deterministic disposal, no orphanable process tree at all.

PR #128, opened 2026-07-13
chore(ui,web): brand cleanups — pixel favicon, lazy tiles, type steps, footer de-dupe

zachdunn:
Small-batch follow-up to #127.

- **Pixel favicon.** The tab icon is now the pixel-grid chevron mark
  (`shape-rendering: crispEdges`, accent `#c27eff`) instead of the smooth-stroke
  version — matches the Geist Pixel wordmark it sits next to.
- **invite page brand font.** `/invite` declared a raw system-mono stack and never
  loaded the brand face; it now uses `var(--mono)` (BaseHead already ships Geist Mono
  there).
- **GalleryTile thumbnails lazy-load.** The thumb was a `background-image` span, so
  every tile fetched immediately regardless of scroll position; it's now a real
  `` with `object-fit: cover` (empty-state
  chevron placeholder unchanged).
- **Type steps on signed-in surfaces.** account/admin/console sizes sprawled across
  10–12.5px in half-pixel increments; consolidated onto 11 / 12 / 14px steps
  (smallest labels raised from 10px, meta folded to 12px, body untouched).
- **Homepage footer de-dupe.** The install section's footnote repeated the `docs`
  and `source & roadmap` links that the site footer renders directly beneath it;
  trimmed the footnote to the access note only.

The Claude Design "Uploads" project was re-synced (render check 12/12 clean; bundle,
styles, and manifest re-uploaded).

Homepage after cleanups

Verified: `@uploads/ui` build + typecheck clean, `astro check` 0 errors, 28 web tests
green, pixel favicon and trimmed footnote confirmed in the browser.


zachdunn:
Also included (per follow-up): the brand mark in the `Brand` lockup is now the same pixel-grid chevron as the favicon — token fills preserved, CSS border-radius dropped since the pixel frame has its own stepped corners. DS re-synced (bundle carries the mark; previews render it in both sizes).

Pixel brand mark in the lockup

zachdunn:
Copy pass folded in: login h1 is just "Sign in" (lockup above carries the name), compact footers drop the leading "uploads.sh ·", the install heading loses "click to copy", the skills command drops `--skill uploads-cli`, and the skill/mcp rows now nest under `uploads install` (which runs both — verified against `install.ts`, default is `all`).

Nested install rows

PR #134, opened 2026-07-13
feat: add authenticated local development stack

zachdunn:
## In plain terms

This adds a single command for the real local authenticated path: Auth, API, and the web app start together, the browser becomes an ordinary `dev-demo` member, and the page shows real workspace data instead of a mocked or unauthenticated state.

- Starts all three services on exact `127.0.0.1` origins, with supervised process-group cleanup.
- Creates the demo identity only behind an ephemeral, development-only Auth gate; it still uses normal Better Auth sessions and API membership checks.
- Seeds two nested PNG fixtures through the normal API and verifies the session, membership, prefix isolation, and file listing with a smoke check.
- Preserves Auth/API outages as an explicit unavailable state rather than treating users as signed out or having no workspaces.
- Does not make simulated local R2 object previews/downloads work at `storage.uploads.sh`.

## How to try it

```bash
pnpm dev:stack
# Open http://127.0.0.1:4321/account/workspaces
pnpm dev:stack:check --json
```

## Technical notes

The local demo endpoint requires the exact stack flag, development environment, loopback Auth/Web origins, and browser Origin. The stack reaps Worker process groups on normal shutdown, startup interruption, unexpected exit, and spawn failure.

## Test plan

- [x] Web client tests: auth timeout/outage handling, workspace outage handling, and local demo session bootstrap (7 tests)
- [x] API session-auth tests (7 tests)
- [x] Scoped formatting, lint, script syntax, and diff checks
- [ ] Full `@uploads/web` typecheck — `astro check` stalled in this local environment after generating Worker types; CI should run the full check.

## Summary by CodeRabbit

* **New Features**
  * Added a one-command authenticated local development stack with readiness checks, smoke tests, demo sessions, and fixture data.
  * Added clearer authentication-unavailable states, retry actions, and timeout handling across account, admin, console, invitation, and device experiences.
  * Workspace loading now distinguishes temporary service outages from empty workspace results.

* **Bug Fixes**
  * Authentication service failures now return a clear temporary-unavailability response instead of appearing signed out.

* **Documentation**
  * Updated local setup, troubleshooting, and operations guidance for the new development workflow.


zachdunn:
@coderabbitai review

PR #143, opened 2026-07-13
feat(uploads): agent-friendly put + install/auth onboarding

zachdunn:
Makes the `uploads` CLI friendlier for the two callers that hit the most friction: agents driving `put` non-interactively, and first-time humans setting it up. Grounded in live agent feedback.

## `put` — agent DX

- **Failures surface on stdout for scripted formats.** With `--format json` (or global `--json`) a failure prints the `{error,code,status}` object to stdout; with `--format url`/`markdown` it prints a one-line `error: …` to stdout too. So `uploads put … --format url 2>/dev/null` is no longer silently blank on failure — the single thing that actually cost the reporting agent a retry. Human/default format is unchanged (stderr only).
- **`--name <leaf>`** overrides just the key's leaf (and default alt) while keeping the stable `--pr`/`--issue` and default-key paths — so an auto-timestamped capture (`wf-hero-2026-…Z.webp`) becomes a clean, stable public URL. Not combinable with `--key`; the `--key`-with-`--pr` error now points at `--name`.
- **`--dry-run`** resolves the object key **and** the exact final public URL without uploading, via a server-side `PUT ?dryRun=1` branch (`apps/api`) that validates key policy and computes the URL from the same `publicUrl` source of truth — no R2 write, no ledger. Lets an agent commit the URL to a PR body before uploading.
- **`FILE_NOT_FOUND`** — distinct error code (exit 2) for a missing local file, so agents branch on `code` instead of parsing messages. A shared `readFileArg` covers both CLI and MCP reads.
- The `put` **MCP tool** mirrors `dryRun`; `filename` doubles as the key-leaf override.

## `install` / auth — onboarding

- Shorter `uploads install` progress (step labels, no child stdout unless `--verbose` or on failure); non-interactive skills install; success next-steps; MCP-without-a-token is skipped with a `uploads login` nudge while the skill still installs.
- Missing-token output reads as onboarding (no `error:` prefix) but still exits non-zero; `--json` keeps `MISSING_TOKEN`. A rejected token stays `UNAUTHORIZED` with a re-login hint.

## Notes

- Exit-code contract: `0` ok · `2` usage/token/file · `3` auth/policy/quota · `4` network · `1` other.
- Server `--dry-run` caveat (commented in code): the key is exact for prefixed keys (what CLI/MCP always send); a bare key re-governs to a fresh `f/<id>/…` on the real upload, and dry-run skips the budget check.
- Docs: `uploads-cli` skill + package README updated. Changeset: `minor` on `@buildinternet/uploads` (the API worker deploys rather than publishes).

## Testing

- `@buildinternet/uploads`: 186 tests pass (added `--name`/`--dry-run`/`FILE_NOT_FOUND` CLI tests, an error-surfacing test, and the concurrent install/auth tests).
- `@uploads/api`: 242 pass (added dry-run route tests: resolves key+URL without writing, rejects a policy-disallowed key).
- Typecheck, lint, and format clean across the repo. Built CLI exercised end-to-end for the error paths.
- Not verifiable end-to-end until deployed: `--dry-run` against prod (the server branch ships with this PR); covered by unit tests on both sides.

### DIFF
--- a/.changeset/uploads-agent-dx.md
+++ b/.changeset/uploads-agent-dx.md
@@ -1,0 +1,15 @@
+---
+"@buildinternet/uploads": minor
+---
+
+CLI onboarding and agent-friendly put:
+
+- **`uploads install`** — short progress, no child stdout unless `--verbose`/failure;
+  non-interactive skills (`-g -y -a '*'`); success next-steps; MCP without a token is
+  skipped with a login nudge (skill still installs).
+- **Missing token** — onboarding copy (no `error:` prefix); exit non-zero; `--json`
+  keeps `MISSING_TOKEN`. Rejected tokens stay `UNAUTHORIZED` with a re-login hint.
+- **`put --name <leaf>`** — clean key leaf on the stable `--pr`/default path.
+- **`put --dry-run`** — resolve key + public URL without writing (API `?dryRun=1`).
+- **Scripted failures** — `--format json|url|markdown` also print on stdout.
+- **`FILE_NOT_FOUND`** — distinct code (exit 2) for a missing local file.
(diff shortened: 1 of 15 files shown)


zachdunn:
@coderabbitai review

PR #155, opened 2026-07-13
fix(web): deliver CSP as HTTP header so frame-ancestors applies

### LINKED ISSUE #153: fix(web): deliver CSP as HTTP header so frame-ancestors applies
## Problem

Browsers log:

> The Content Security Policy directive `frame-ancestors` is ignored when delivered via a `<meta>` element.

Several signed-in and auth surfaces on `apps/web` set CSP only via:

```html
<meta http-equiv="Content-Security-Policy" content="…; frame-ancestors 'none'; …">
```

Examples today:

- `AccountLayout.astro` / `AdminLayout.astro` (`signedInCsp`)
- `login.astro`, `device.astro`, `accept-invitation/[id].astro`
- `invite.astro` (also has a matching header in `public/_headers` for `/invite*`)

Public file/gallery pages already set CSP on the **response header** (`applyPublicFileHeaders` / gallery helpers), so they are fine for `frame-ancestors`.

`frame-ancestors` only applies when the policy is delivered as an **HTTP response header** (or the older `X-Frame-Options`). Meta CSP cannot enforce clickjacking protection.

## Goal

Ship CSP (including `frame-ancestors 'none'`) as a **response header** for the signed-in / auth pages that currently rely on meta-only CSP, without weakening the rest of the policy and without double-policy footguns (multiple CSPs all must pass).

## Suggested approach

1. Prefer a single authoritative CSP **header** for `/account/*`, `/admin/*`, login, device, accept-invitation (and audit invite: keep header + meta in sync or drop redundant meta).
2. Options:
   - Cloudflare Pages/`_headers` patterns where origins are fixed, **or**
   - Worker/middleware on `apps/web` that sets `Content-Security-Policy` from the same builders as today (`signedInCsp`, page-local CSP strings).
3. Keep `CF_RUM_*` fragments in `src/lib/csp.ts` as the source of truth (including `'self'` for `/cdn-cgi/rum`).
4. Avoid stacking divergent meta + header policies unless intentionally identical.
5. Tests: assert header presence for the signed-in shell paths (or unit-test the middleware/`_headers` mapping).

## Out of scope

- Changing the actual CSP allowlist beyond delivering it correctly as a header.
- RUM `connect-src` (already fixed to include `'self'` for first-party `/cdn-cgi/rum`).

## References

- [MDN: CSP frame-ancestors](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/frame-ancestors) — not valid in meta
- Internal: `apps/web/src/lib/signed-in-page.ts`, `apps/web/src/lib/csp.ts`, `apps/web/public/_headers`

zachdunn:
## In plain terms

Browsers were ignoring clickjacking protection (`frame-ancestors`) because CSP was only set in a `<meta>` tag. Account, admin, and auth pages now send CSP as a real HTTP header, and they get the same security-header baseline as public file/gallery pages (not just CSP).

## What it does

- Sets **CSP via `Astro.response.headers`** (not meta) on `/account/*`, `/admin/*`, login, device, accept-invitation
- Shared helper **`applyAuthSecurityHeaders`**: CSP + `Referrer-Policy`, `nosniff`, `X-Frame-Options`, `COOP`, `Permissions-Policy`, `noindex`, `no-store` — same suite as public file/gallery
- `/invite*`: header-only CSP via `public/_headers` (synced to `INVITE_CSP`), plus matching frame/COOP headers
- Removes meta CSP to avoid stacked divergent policies

## What it is not

- Does not change CSP allowlists (RUM/`'self'` already fixed elsewhere)
- Does not cover HSTS (Cloudflare zone) or marketing-page baseline — see #156

## Technical notes

- Builders: `signedInCsp`, `authPageCsp`, `INVITE_CSP` in `apps/web/src/lib/signed-in-page.ts`
- Unit tests assert header suite + `_headers` ↔ `INVITE_CSP` equality

## Test plan

- [x] `pnpm --filter @uploads/web test` (78 tests)
- [ ] Spot-check production response headers after deploy (`curl -sI` account/login)

Closes #153

## Summary by CodeRabbit

* **Security Enhancements**
  * Authentication, invitation, device, and account pages now apply Content Security Policy and related protections through response headers.
  * Invitation pages enforce consistent security policies, including framing restrictions, referrer controls, caching rules, and crawler protections.
  * Removed HTML-based CSP handling in favor of more reliable server-delivered security headers.

* **Tests**
  * Added coverage to verify security policies and required response headers across authentication and invitation flows.

### DIFF
--- a/apps/web/public/_headers
+++ b/apps/web/public/_headers
@@ -17,6 +17,8 @@
   Referrer-Policy: no-referrer
   X-Robots-Tag: noindex, nofollow, noarchive
   X-Content-Type-Options: nosniff
+  X-Frame-Options: DENY
+  Cross-Origin-Opener-Policy: same-origin
   Content-Security-Policy: default-src 'none'; connect-src https://api.uploads.sh 'self' https://cloudflareinsights.com; script-src 'self' 'unsafe-inline' https://static.cloudflareinsights.com; style-src 'self' 'unsafe-inline'; font-src 'self'; img-src data:; base-uri 'none'; form-action 'none'; frame-ancestors 'none'
   Permissions-Policy: camera=(), microphone=(), geolocation=()
 
--- a/apps/web/public/_headers
+++ b/apps/web/public/_headers
@@ -10,13 +10,14 @@
   Link: </sitemap.xml>; rel="sitemap"; type="application/xml"
   X-Content-Type-Options: nosniff
 
-# Keep CSP in sync with src/pages/invite.astro (header + meta both apply).
+# Keep CSP identical to INVITE_CSP in src/lib/signed-in-page.ts (header only —
+# meta CSP is not used; frame-ancestors requires an HTTP header).
 /invite*
   Cache-Control: no-store
   Referrer-Policy: no-referrer
   X-Robots-Tag: noindex, nofollow, noarchive
   X-Content-Type-Options: nosniff
-  Content-Security-Policy: default-src 'none'; connect-src 'self' https://api.uploads.sh https://cloudflareinsights.com; script-src 'self' 'unsafe-inline' https://static.cloudflareinsights.com; style-src 'self' 'unsafe-inline'; font-src 'self'; img-src data:; base-uri 'none'; form-action 'none'; frame-ancestors 'none'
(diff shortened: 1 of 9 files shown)


zachdunn:
@coderabbitai review

PR #173, opened 2026-07-15
feat: human-size optimize notes and report put overwrites

zachdunn:
## In plain terms

The CLI now prints readable optimize sizes (e.g. `411.5 KB → 94.2 KB`) and tells you when a re-upload overwrote an existing object instead of creating a new one. Overwrites stay silent on purpose — no opt-in prompt — so stable PR attachment URLs keep hot-swapping for agents. Docs cover custom metadata examples (`url`, `path`, `app`) and re-upload behavior.

## What it does

- API `PUT` responses include `replaced: true|false` (derived from the pre-upload head already used for metering; files-sdk has no size formatter or replaced flag on `UploadResult`)
- CLI human mode: `>> replaced existing object (same URL)` when overwriting
- Optimize notes use human sizes; homepage terminal mock matches
- Docs/skill/README: hot-swap semantics + metadata examples for screenshot context
- Changeset for `@buildinternet/uploads` (patch)

## What it is not

- Not a confirmation/`--force` gate on overwrite (would break agent re-runs)
- Does not change public URL behavior or embed cache policy

## How to try it

```bash
uploads attach ./before.png
# re-run the same filename on the same PR:
uploads attach ./before.png
# expect: >> replaced existing object (same URL)
# optimize line like: >> optimized 411.5 KB → 94.2 KB (before.webp)
```

## Technical notes

- files-sdk: `exists()` / `head().size` only — no `formatBytes`; `UploadResult` is `{ key, size, contentType, etag?, lastModified? }`
- `putObject` already HEADs for ledger delta; `replaced` is that signal surfaced to clients
- `formatByteSize` lives in `@buildinternet/uploads` (CLI); UI keeps its own formatter

## Test plan

- [x] `pnpm --filter @buildinternet/uploads test` (includes format-bytes)
- [x] `pnpm --filter @uploads/api test` (replaced flag on re-PUT)
- [x] package typecheck for uploads + api
- [ ] Spot-check homepage terminal mock locally if deploying web

### DIFF
--- a/.changeset/human-sizes-replaced.md
+++ b/.changeset/human-sizes-replaced.md
@@ -1,0 +1,5 @@
+---
+"@buildinternet/uploads": patch
+---
+
+Print human-readable optimize sizes (e.g. `411.5 KB → 94.2 KB`) and note when a put overwrites an existing object (`replaced` on the API/JSON, `>> replaced existing object (same URL)` in human mode).
--- a/.changeset/human-sizes-replaced.md
+++ b/.changeset/human-sizes-replaced.md
@@ -2,4 +2,4 @@
 "@buildinternet/uploads": patch
 ---
 
-Print human-readable optimize sizes (e.g. `411.5 KB → 94.2 KB`) and note when a put overwrites an existing object (`replaced` on the API/JSON, `>> replaced existing object (same URL)` in human mode).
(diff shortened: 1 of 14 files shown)


zachdunn:
Follow-ups landed on this branch:

- **Dry-run now reports would-replace** when the key already exists (`>> would replace existing object (same URL)` / `"replaced": true` without writing).
- **Follow-up issue** for opt-in/opt-out overwrite policy: https://github.com/buildinternet/uploads/issues/174

Still true: real puts overwrite with **no pre-confirmation** — we only tell you after (or on dry-run before you choose to proceed).

PR #182, opened 2026-07-15
feat(web): default the invite page to the code-stdin login flow

zachdunn:
Last of the onboarding-walkthrough hygiene items (after #176/#177/#179/#181). The /invite page baked the one-time `upe_` enrollment code into the copyable login command, so redeeming an invite put a password-equivalent secret into the user's shell history.

## What changed

When the page has a code in the URL fragment, the login step is now the secret-free `uploads login --code-stdin`, and the code becomes its own copyable step ("Paste your code at the prompt"). The CLI already reads the code from stdin in this mode, so nothing secret ever appears on a command line. The security note now explains why.

Note on the alternative: a `printf '%s' upe_… | uploads login --code-stdin` one-liner was rejected because the code would still land in shell history — the paste-at-prompt flow is the only genuinely history-safe shape.

No behavior change when the fragment carries no code (the plain `uploads login` row stays). Fragment scrubbing from the address bar is unchanged.

## Verified

Loaded /invite with a synthetic fragment code in the dev server: login row reads `uploads login --code-stdin` with matching copy payload, the code step is visible with the code as its copy payload, and `location.hash` is scrubbed.

### DIFF
--- a/apps/web/src/pages/invite.astro
+++ b/apps/web/src/pages/invite.astro
@@ -25,9 +25,10 @@ import Footer from "../components/Footer.astro";
     <ol>
       <li>Install<div class="command"><code>npm install --global @buildinternet/uploads</code><button data-copy="npm install --global @buildinternet/uploads" aria-live="polite">copy</button></div></li>
       <li>Log in<div class="command"><code id="login-cmd">uploads login</code><button id="login-copy" data-copy="uploads login" aria-live="polite">copy</button></div></li>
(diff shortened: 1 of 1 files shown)


zachdunn:
Closing unmerged: the code is single-use and atomically redeemed, so a spent code in shell history is worthless. The only residual window (a failed login leaving a live code) is bounded by code expiry and dominated by local-machine compromise, where the saved 90-day workspace token is the bigger prize anyway. The two-copy flow isn't worth the friction on a first-run experience; `--code-stdin` remains documented for anyone who wants it.

PR #226, opened 2026-07-18
feat: OAuth 2.1 authorization server for the hosted MCP

### LINKED ISSUE #224: feat(mcp): OAuth authorization server for agents.uploads.sh so the remote MCP authenticates without a baked token
## Problem

The remote MCP server at `https://agents.uploads.sh/mcp` authenticates with a
workspace bearer token (`Authorization: Bearer up_<workspace>_…`). That works
for hand-wired clients, but it means the hosted endpoint **cannot authenticate
itself** in a distributable Claude Code plugin:

- `uploads login` stores the token in a **config file**
  (`~/.config/buildinternet/config`), not an environment variable
  (`packages/uploads/src/config-file.ts`, `src/commands/login.ts`).
- A static plugin `.mcp.json` can only interpolate `${UPLOADS_TOKEN}` from the
  **environment**, which a normally signed-in user won't have set → the hosted
  MCP 401s.
- `uploads install` sidesteps this by reading the token from config and
  **baking the literal value** into `claude mcp add … --header` at install
  time. That's per-user and per-machine — it can't ship inside a plugin that
  other people install.

### Current workaround in the plugin

The Claude Code plugin (`.claude-plugin/marketplace.json`,
`plugins/claude/uploads/.mcp.json`) bundles the **local stdio** server
(`uploads mcp`), which reads the same config file and works after
`uploads login`. The tradeoff: it requires the `uploads` CLI on `PATH` and a
prior interactive login, so it doesn't "just work" for a user who installs the
plugin without first setting up the CLI.

## Why OAuth

We already publish RFC 9728 `oauth-protected-resource` metadata on the `api`
and `agents` origins (#216), but there is **no authorization server**, so
Claude Code's remote-MCP client can't complete an interactive flow — the
metadata advertises a resource with nowhere to authenticate.

Standing up a real OAuth 2.1 authorization server for `agents.uploads.sh` would
let the hosted MCP authenticate through the browser with **no env var, no baked
token, and no local CLI dependency** — the bundled plugin could then point at
the hosted endpoint and work for any user.

## Proposal

- Expose an OAuth 2.1 authorization server for `agents.uploads.sh` (reusing the
  existing dedicated auth worker / better-auth stack where possible).
- Support the pieces the MCP client expects: RFC 8414 AS metadata, PKCE, and
  ideally RFC 7591 Dynamic Client Registration so we don't hand-register each
  client.
- Mint the same workspace-scoped tokens the bearer scheme already issues
  (`files:read` / `files:write`), so downstream authorization is unchanged.
- Decide initial blast radius: first-party plugin only vs. general third-party
  clients (the CLI device flow already uses a fixed client id — this adds a
  general AS).

## Acceptance

A fresh Claude Code user can `/plugin install uploads@uploads`, and the hosted
MCP at `agents.uploads.sh/mcp` authenticates via a browser OAuth flow with no
manual token steps and without the `uploads` CLI installed. Once that lands,
switch the plugin's bundled `.mcp.json` back to the hosted HTTP endpoint.

## References

- `apps/mcp` — remote MCP worker
- `https://uploads.sh/auth.md` — "What we deliberately do not publish" (no
  public OAuth AS today)
- #216 — RFC 9728 protected-resource metadata (the discovery half already ships)
- `packages/uploads/src/config-file.ts` / `src/commands/login.ts` — token is
  stored in config, not env
- `plugins/claude/uploads/.mcp.json` — current stdio workaround

zachdunn:
Closes #224.

Stands up an OAuth 2.1 authorization server on the existing `uploads-auth` worker so the hosted MCP at `agents.uploads.sh/mcp` authenticates through a browser flow — no env var, no baked token, no local CLI dependency. The bundled Claude Code plugin points back at the hosted HTTP endpoint.

Design doc: `docs/superpowers/specs/2026-07-17-oauth-authorization-server-design.md`. Pattern follows the proven sibling implementations (releases.sh, sunny) on Better Auth 1.6.23 + `@better-auth/oauth-provider`.

## What's in here

**apps/auth — the authorization server**
- `jwt()` + `oauthProvider()` added to the existing `betterAuth()` config (jwt first — it signs access tokens and serves JWKS at `/api/auth/jwks`). Issuer: `https://auth.uploads.sh/api/auth`.
- Five new D1 tables (`oauth_client`, `oauth_access_token`, `oauth_refresh_token`, `oauth_consent`, `jwks`) as hand-synced schema + migration `20260717000000_oauth_provider.sql`.
- `customAccessTokenClaims` embeds workspace claims from org memberships (`workspace`: oldest membership's slug, `workspaces`: all slugs). Zero-workspace users still get a token; the resource server owns the 403.
- Dynamic client registration on (unauthenticated, 5/min rate limit) with a stale-client reaper wired into the existing cron (default observe-only via `OAUTH_CLIENT_REAPER_ENABLED`).
- RFC 8414 discovery aliases at the issuer root (`/.well-known/oauth-authorization-server[/*]`), CORS `*`.
- Scopes are the existing `FILE_SCOPES` (`files:read` / `files:write` / `files:delete`); DCR default grants read+write.

**apps/mcp — the resource server**
- JWT-shaped bearers verify via `jose` against the AS JWKS (5-min in-isolate cache, no service binding); non-JWT bearers use the existing `up_` path untouched.
- Valid token → same context (`workspace`/`authScopes`) as today, so tools are unchanged downstream. `workspace: null` → actionable 403. Bad/expired token → 401 + `WWW-Authenticate: Bearer error="invalid_token", resource_metadata=…`; **no credential at all** → 401 + the RFC 9728 discovery challenge (this is the response MCP clients bootstrap OAuth from).
- `oauth-protected-resource` metadata now advertises `authorization_servers: ["https://auth.uploads.sh/api/auth"]` (MCP origins only — `api.uploads.sh` keeps omitting it since it doesn't accept these JWTs yet).

**apps/web — the human surface**
- New `/oauth/consent` page (Astro, dependency-free, `authPageCsp`, all `textContent` sinks, sanitized client URIs).
- `auth-client.ts` injects `oauth_query` into sign-in POSTs when the page carries the AS's signed authorize query, which is how the flow resumes through login (verified against the plugin source).
- `auth.md` / README / llms.txt updated — the "no public OAuth AS" claim is gone.

**plugins/claude** — `.mcp.json` switches from the local stdio workaround back to `{"type":"http","url":"https://agents.uploads.sh/mcp"}`.

## Notes

- Access tokens are JWTs **only when the client sends an RFC 8707 `resource` param** matching `validAudiences` (plugin behavior); MCP clients per the 2025-06 spec always do. A client that omits it gets an opaque token the MCP worker rejects — acceptable for v1.
- `openid-configuration` correctly 404s: this AS issues no OIDC scopes.
- Multi-workspace users get their oldest workspace in v1; all slugs ride in the token for a future consent-time picker (deferred, see design doc).

## Verification

- 1278/1278 tests pass repo-wide, including new suites: auth (DCR, JWKS, discovery aliases, workspace claims, reaper), mcp (signed test JWTs: happy path, both `scope` shapes, bad iss/aud/exp, deleted workspace, no-workspace 403, no-credential challenge, `up_` path regression), web (oauth_query injection).
- Typecheck clean for auth/api; the 2 mcp errors and 1 web error reproduce identically on `main` (pre-existing `worker-configuration.d.ts` drift, spun off separately).

## Post-merge ops

1. `wrangler d1 migrations apply DB --remote` in `apps/auth` (uploads-auth database).
2. Deploy `uploads-auth`, `uploads-mcp`, `uploads-web`.
3. End-to-end prod check: fresh `claude mcp add --transport http uploads https://agents.uploads.sh/mcp` → browser consent → tool call; then re-run the discovery scan from #216.

## Summary by CodeRabbit

* **New Features**
  * Added OAuth 2.1 authorization for the hosted MCP server, including JWT bearer authentication, scopes, workspace claims, discovery, dynamic client registration, and consent approval.
  * Added OAuth login-resume support for GitHub and magic-link authentication.
  * Hosted MCP integrations now connect through `https://agents.uploads.sh/mcp` with interactive browser authorization.

* **Bug Fixes**
  * Improved OAuth metadata and authentication error responses, including standards-compliant discovery challenges.

* **Documentation**
  * Updated agent authentication, OAuth, and hosted MCP setup guidance.

### DIFF
--- a/apps/api/src/well-known.test.ts
+++ b/apps/api/src/well-known.test.ts
@@ -28,7 +28,9 @@ describe("GET /.well-known/oauth-protected-resource", () => {
     expect(body.scopes_supported).toEqual(["files:read", "files:write", "files:delete"]);
     expect(body.bearer_methods_supported).toEqual(["header"]);
     expect(body.resource_documentation).toBe("https://uploads.sh/auth.md");
-    // No public authorization server for third-party clients (see well-known.ts).
+    // apps/api's own call site deliberately doesn't pass `authorizationServers`
+    // (see well-known.ts) — only apps/mcp does (issue #224). Not that the
+    // field is unsupported; index.ts just never opts in.
     expect(body.authorization_servers).toBeUndefined();
   });
 });
--- a/apps/api/src/well-known.ts
+++ b/apps/api/src/well-known.ts
@@ -7,14 +7,16 @@
  * programmatically. It is served at `/.well-known/oauth-protected-resource`
  * on each resource origin (api.uploads.sh, agents.uploads.sh).
  *
- * `authorization_servers` is deliberately omitted. There is no public OAuth
- * authorization server for third-party clients — tokens are minted out-of-band
- * via `uploads login` (a first-party device flow) and the token-mint endpoint,
- * all described in `resource_documentation` (auth.md). Advertising an
- * `authorization_servers` entry would point RFC 9728 clients at an
- * authorization-server metadata document that we intentionally do not publish,
- * so we leave it out rather than dangle a broken reference. See auth.md's
- * "What we deliberately do not publish".
+ * `authorization_servers` is opt-in per caller via `authorizationServers`
+ * (issue #224). Only apps/mcp passes it — it's the only resource server that
+ * verifies the uploads-auth OAuth JWTs (v1 is MCP-only; see
+ * docs/superpowers/specs/2026-07-17-oauth-authorization-server-design.md).
+ * apps/api's own usage keeps omitting it: tokens for the REST API are still
+ * minted out-of-band via `uploads login` / the token-mint endpoint (described
+ * in `resource_documentation`), and advertising an authorization server here
+ * would point RFC 9728 clients at an AS that doesn't accept api.uploads.sh
+ * audiences — a dangling, misleading reference. See auth.md's "What we
+ * deliberately do not publish".
  */
 import { FILE_SCOPES } from "./auth-db";
 
@@ -28,12 +30,16 @@ export interface ProtectedResourceMetadata {
   bearer_methods_supported: string[];
   /** Human/agent-readable acquisition + usage docs (auth.md). */
   resource_documentation: string;
+  /** Issuer URL(s) of the authorization server(s) that issue tokens for this resource. Omitted when the caller doesn't pass `authorizationServers`. */
+  authorization_servers?: string[];
 }
 
 export function protectedResourceMetadata(opts: {
   resource: string;
   resourceName: string;
   webOrigin: string;
+  /** Issuer(s) whose access tokens this resource server accepts. Leave unset to omit `authorization_servers` (the honest default — see module doc). */
+  authorizationServers?: string[];
 }): ProtectedResourceMetadata {
   return {
     resource: opts.resource,
@@ -41,6 +47,7 @@ export function protectedResourceMetadata(opts: {
     scopes_supported: [...FILE_SCOPES],
     bearer_methods_supported: ["header"],
     resource_documentation: `${opts.webOrigin.replace(/\/$/, "")}/auth.md`,
+    ...(opts.authorizationServers ? { authorization_servers: opts.authorizationServers } : {}),
   };
 }
 
--- a/apps/auth/migrations/20260717000000_oauth_provider.sql
+++ b/apps/auth/migrations/20260717000000_oauth_provider.sql
@@ -1,0 +1,93 @@
+-- OAuth 2.1 authorization server (issue #224, Lane A): `jwt()` +
+-- `oauthProvider()` plugins (see src/auth.ts). Paired with `jwks` /
+-- oauthClient / oauthAccessToken / oauthRefreshToken / oauthConsent in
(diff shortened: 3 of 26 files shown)


zachdunn:
@coderabbitai Addressed in 32758c2:

- **auth.ts ordering** — added `member.id` secondary sort for deterministic primary-workspace selection.
- **reaper atomicity** — restructured: delete now runs a single statement with correlated `NOT EXISTS` checks against consent/access-token/refresh-token tables plus the age/ownership/trust predicates re-evaluated at statement time; no more collect-then-delete read, and no full usage-table scans into memory.
- **RFC 8414 test** — now exercises the real issuer-derived path `/.well-known/oauth-authorization-server/api/auth`.
- **bearer scheme** — parsed case-insensitively per RFC 9110, with a regression test.
- **consent deny** — now follows `redirect_uri` so the client receives `error=access_denied` + `state`; the dead-end local panel is gone.
- **stubLocation** — hoisted to module scope.

Not applied: the mcp.test.ts duplicate-declaration finding is a false positive — there is a single `let oauthJwks` declaration (line 50), and the two `const body` bindings live in separate test-function scopes. The file typechecks and all 55 tests pass.

Issue #232, opened 2026-07-18
chore: typecheck is red on main (apps/mcp + apps/web)


zachdunn:
Three pre-existing errors fail `pnpm typecheck` on main, all worker-configuration.d.ts ambient-type drift, none from the OAuth PRs (verified identical on main before #226):

- `apps/api/src/guards.ts:227/239` (surfaces in **apps/mcp**'s typecheck, which compiles api sources cross-package): `RENDER_LIMITER`/`WS_CREATE_LIMITER` not on the mcp worker's `Env` — the api-side guards assume api bindings that the mcp-generated Env lacks.
- `apps/web/src/pages/account/workspaces/new.astro` (~:163): HTMLRewriter's ambient `Element.append()` clashes with DOM `ParentNode` — `bindCopyButtons(requireElement<HTMLElement>(…))`. The consent page dodged the same trap with `appendChild`; same class of fix applies (or narrow the ambient types).

Green typecheck on main would let CI enforce it and stop these masking real regressions.

PR #236, opened 2026-07-18
feat(auth,web): workspace picker at OAuth consent for multi-workspace users

### LINKED ISSUE #231: feat(web): workspace picker at consent for multi-workspace users
Deferred from the #224 design (docs/superpowers/specs/2026-07-17-oauth-authorization-server-design.md): OAuth access tokens embed `workspace` = the user's **oldest** membership, with all slugs in a `workspaces` claim. Multi-workspace users can't choose which workspace an OAuth-connected MCP client operates on — the oldest silently wins.

Add a workspace selector to /oauth/consent when the signed-in user has >1 organization (the page already fetches `listOrganizations`; #229 added the zero-org interstitial). Needs a way to thread the choice into token claims — likely a consent-time parameter the AS's `customAccessTokenClaims` can read, or scoping the consent record. Check what better-auth's oauth-provider supports before designing.

zachdunn:
Closes #231.

Multi-workspace users can now choose which workspace an OAuth grant operates on, instead of the oldest membership silently winning.

## How it works

The mechanism is the oauth-provider plugin's `postLogin.consentReferenceId` hook (verified against the 1.6.23 dist):

- The consent page shows a **workspace select** when the user has 2+ org memberships (hidden otherwise; default = oldest). On Allow it first stores the choice via the new session-authed `POST /api/auth/oauth2/workspace-choice`, then submits consent.
- The hook returns `ws:<slug>` (stored choice if still a live membership, else oldest) for multi-workspace users, `undefined` for 0/1 memberships. The plugin keys `oauth_consent` rows by this string at both authorize- and consent-time, and threads it through the authorization code into access/refresh token rows.
- `customAccessTokenClaims` receives it back as `referenceId` at every issuance and overrides the `workspace` claim when the slug is still one of the user's `workspaces` — per-grant semantics, refresh tokens keep the originally granted workspace.

Properties that fall out of the consent-record keying:
- **Changing the choice re-triggers the consent screen** (no stale-grant dead end).
- **Single-workspace users and existing grants are completely untouched** (`referenceId` stays null — no re-consent churn on deploy).
- **Existing multi-workspace users re-consent once**, which is exactly when they first see the picker.

## Changes

- `apps/auth`: `oauth_workspace_choice` table + migration; `workspace-choice.ts` (endpoint plugin + `resolveWorkspaceChoiceReferenceId` + `applyWorkspaceChoice`, all exported for direct testing); wiring in `auth.ts`. 14 new tests (fake-D1).
- `apps/web`: picker UI in `consent.astro` (static markup, dynamic options via `appendChild` per the HTMLRewriter ambient-type convention); `setOAuthWorkspaceChoice` in `auth-client.ts`; `orderOrgsOldestFirst` in `workspace-ui.ts`. 8 new tests.

## Notes for review

- The picker's default ordering uses the **organization's** `createdAt` (what `/organization/list` exposes) while the AS default uses **membership** age. They can differ for invited users, but once the picker is visible the choice is explicit and posted, so this only affects which option is pre-selected.
- Choice-endpoint failures keep the user on the consent panel with an error rather than submitting consent for the wrong workspace.
- No changeset (apps/* are ignored packages).

## Testing

- `pnpm test`: 119 files / 1305 tests pass.
- `pnpm check` clean; auth typecheck clean; web typecheck has only the pre-existing ambient-type errors tracked in #232.

## Post-merge

- CI applies the D1 migration on deploy (same path as #226's).
- Prod verification: authorize with the dunn.zach@gmail.com multi-workspace user → picker appears → choose non-oldest → token's `workspace` claim reflects it.

## Summary by CodeRabbit

* **New Features**
  * Added workspace selection to the OAuth consent flow for users with multiple workspaces.
  * Remembers the selected workspace for future OAuth authorizations.
  * Applies the selected workspace to issued OAuth access tokens.
  * Added validation to prevent selecting workspaces the user cannot access.
  * Displays workspaces in oldest-first order and provides clear feedback if saving a selection fails.

* **Bug Fixes**
  * Invalid or outdated workspace selections now safely fall back to an available workspace.

### DIFF
--- a/apps/auth/migrations/20260718000000_oauth_workspace_choice.sql
+++ b/apps/auth/migrations/20260718000000_oauth_workspace_choice.sql
@@ -1,0 +1,14 @@
+-- Per-grant workspace choice (issue #231, auth side): stores a
+-- multi-workspace user's last explicit workspace pick for the OAuth 2.1
+-- authorization server. One row per user, keyed by user_id. Not a
+-- @better-auth/oauth-provider table -- read/written by
+-- src/workspace-choice.ts (see src/schema.ts:oauthWorkspaceChoice). Timestamps
+-- are integer epoch ms, matching every other hand-synced table in this
+-- schema (see migrations/20260717000000_oauth_provider.sql).
+
+CREATE TABLE oauth_workspace_choice (
+  user_id TEXT PRIMARY KEY,
+  workspace TEXT NOT NULL,
+  created_at INTEGER NOT NULL,
+  updated_at INTEGER NOT NULL
+);
--- a/apps/auth/src/auth.ts
+++ b/apps/auth/src/auth.ts
@@ -24,6 +24,11 @@ import { localDemoEnabled, localDemoPlugin } from "./local-demo";
 import * as schema from "./schema";
 import { authTrustedOrigins, isTrustedOrigin } from "./trusted-origins";
 import {
+  applyWorkspaceChoice,
+  resolveWorkspaceChoiceReferenceId,
+  workspaceChoicePlugin,
+} from "./workspace-choice";
+import {
   resolveDashApiKey,
   resolveGitHubCredentials,
   resolveSigningSecret,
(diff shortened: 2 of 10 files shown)


zachdunn:
Review findings addressed in 9f0e808:

**Picker default (consent.astro / workspace-ui.ts) — valid, fixed.** The finding was right that this was more than cosmetic: since the page always POSTs the selection on Allow, a client-side default could overwrite a stored choice or shift the token's workspace for a user who touched nothing. Fix: new session-authed `GET /oauth2/workspace-choice` returns the AS's own resolution (stored choice if still a live membership, else oldest *membership*), and the picker defaults its selection to that. Org-`createdAt` ordering remains as display order only, with the doc comment reworded to stop implying it's a membership-age proxy. Covered by 4 new endpoint tests and 2 new client-helper tests (1311 total passing).

**Per-client/grant choice scoping (migration / schema) — skipped, infeasible with this plugin version.** `postLogin.consentReferenceId` receives only `{ user, session, scopes }` — no client id and no request context (verified against the 1.6.23 dist) — so a composite-keyed row could never be read back at authorize-time, and the consent endpoint's body schema is plugin-owned, so the selection can't ride the consent POST either. Per-grant scoping already holds where it matters: the returned `ws:<slug>` is persisted on the consent row and baked into that grant's token rows, so concurrent grants don't share state after consent. The residual is a same-user race across two simultaneously open consent tabs (seconds-wide, recoverable via re-consent); now documented on `effectiveWorkspaceSlug` in workspace-choice.ts.

zachdunn:
Prod e2e completed 2026-07-18, all paths verified live:

- Picker rendered for a 2-workspace account, server-resolved default shown (`buildinternet` on first grant, the stored choice on later grants — the GET-default fix in action).
- Choosing the non-oldest workspace minted a JWT with `workspace: "picker-e2e-tmp"` (`workspaces` carried both), accepted by agents.uploads.sh/mcp.
- Gaining a second membership forced re-consent on an existing null-referenceId grant, as designed.
- Single-workspace accounts see no picker and behave exactly as before.

Migration was CI-applied (`oauth_workspace_choice` live in prod D1). Test client/grants/tokens deleted afterward. The e2e also surfaced two unrelated prod bugs, fixed in #238.

PR #238, opened 2026-07-18
fix(api,web): unbreak self-serve workspace CORS + magic-link consent resume

zachdunn:
Two prod bugs surfaced by the live e2e of the #236 workspace picker.

## 1. Self-serve workspace creation from the console has been broken since #171

`POST /v1/workspaces` authenticates by **session cookie**, but sat under the uncredentialed `consoleCors` mounted on all of `/v1/*`. The preflight response therefore lacked `Access-Control-Allow-Credentials`, the browser dropped the request (`Failed to fetch`), and the create-workspace form showed "The API is unreachable right now" for every signed-in user on uploads.sh. (#171's owed prod signup-flow verification would have caught this — it just got caught now instead, live.)

Fix: `/v1/workspaces` gets the credentialed `adminUiCors` treatment via a conditional mount; every other `/v1/*` route stays uncredentialed (bearer tokens are the boundary there). New `cors.test.ts` pins the three cases (workspaces credentialed, tokens not, /me still credentialed).

## 2. Magic-link OAuth resume fails `invalid_signature` for most sig values

better-auth 1.6.23's magic-link verify decodes its `callbackURL` one time too many. A signed OAuth query round-tripping through a sign-in email (the #229 resume path) lands on `/oauth/consent` with the `sig` param's base64 `%2B` collapsed to a literal `+` — query parsing folds that into a space and the AS rejects the consent POST with `invalid_signature`. Any sig containing `+` (~50% of HMACs) breaks; `%2F`→`/` and `%3D`→`=` survive parsing losslessly, as do the other signed params (URLs and base64url values contain no encoded `+`).

Fix: `repairOAuthQuery` in auth-client.ts re-encodes literal `+` **inside the sig param only** before the consent POST — a no-op for cleanly encoded queries, and independent of which hop dropped the encoding (so it stays correct if better-auth fixes the double-decode later). Unit tests cover the corrupted and clean cases.

Verified locally: 121 files / 1323 tests pass, format clean.

Post-merge: re-run the live flow — self-serve create from uploads.sh, and a fresh magic-link → consent → Allow round trip.

### DIFF
--- a/apps/api/src/cors.test.ts
+++ b/apps/api/src/cors.test.ts
@@ -1,0 +1,47 @@
+/**
+ * CORS mounting on the root app. The subtle one is `/v1/workspaces`: it is
+ * the only `/v1/*` route authenticated by session COOKIE (self-serve creation
+ * from the signed-in console at WEB_ORIGIN), so its preflight must be
+ * credentialed — without `Access-Control-Allow-Credentials` the browser
+ * drops the request entirely ("Failed to fetch") and self-serve creation
+ * silently breaks in prod. The rest of `/v1/*` is bearer-token-authenticated
+ * and deliberately stays uncredentialed.
+ */
+import { describe, expect, it } from "vitest";
+import { app } from "./index";
+
+const env = {} as unknown as Env;
+
+function preflight(path: string) {
+  return app.request(
+    `https://api.uploads.sh${path}`,
+    {
+      method: "OPTIONS",
+      headers: {
+        Origin: "https://uploads.sh",
+        "Access-Control-Request-Method": "POST",
+        "Access-Control-Request-Headers": "content-type",
+      },
+    },
+    env,
+  );
+}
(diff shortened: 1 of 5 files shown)


zachdunn:
Prod-verified 2026-07-18:

- **CORS fix**: self-serve workspace creation from uploads.sh succeeded through the real UI on first try post-deploy (previously every attempt died at preflight with "API unreachable").
- **Sig repair**: a fresh magic-link → consent → Allow round trip completed, and — since that link's sig happened to contain no `+` — the consent page was additionally replayed with a deterministically corrupted signed query (`sig` holding a literal `+`, byte-for-byte what better-auth's double-decode produces). Allow issued an authorization code; pre-fix that exact input failed with `invalid_signature`.

Workaround-removal tracking for the upstream better-auth double-decode is being filed as a separate issue.

PR #253, opened 2026-07-18
feat(auth): migrate CLI device-flow login to a managed official OAuth client (#251)

### LINKED ISSUE #251: Migrate CLI device-flow login to a managed official OAuth client registration
## Context

PR #245 shipped the operator admin panel for OAuth client registrations (`/admin/oauth`), giving us managed, inspectable client records with an `official` flag, disable toggle, and consent/token stats.

The CLI's device-flow login still uses a static, hardcoded client id (`UPLOADS_CLI_CLIENT_ID = "uploads-cli"`) that never touches the `oauth_client` table. That means:

- It doesn't appear in the admin panel — no visibility into its consent/token activity alongside other clients.
- It can't be disabled, rotated, or annotated like managed clients.
- It can't carry the `official` badge/metadata that first-party apps are supposed to have.

## Proposal

Migrate the CLI to a managed registration:

1. Register an official client via the admin panel (public PKCE, `official: true`, appropriate scopes) — or seed it deterministically in a migration so all environments agree on the client id.
2. Point the CLI's device flow at that client id (keeping `uploads-cli` as the id is fine if we seed it — avoids breaking already-configured CLIs).
3. Verify the device-authorization grant path works against a registered client (the AS currently special-cases the static id — check what assumptions exist around it).
4. Remove the static-id special case once the registered client is live.

## Notes

- Deliberately NOT in scope: the hosted MCP server's clients — those stay dynamic-registration (RFC 7591) per the MCP authorization model; DCR + consent + PKCE + the nightly reaper is the intended posture there.
- Backward compatibility for existing CLI logins (refresh tokens minted under the static id) needs a look.

zachdunn:
Closes #251.

## What

The CLI's device-flow login no longer relies on a string-allowlisted static client id. `uploads-cli` is now a managed `oauth_client` registration:

- **Seed migration** `apps/auth/migrations/20260719000000_seed_cli_oauth_client.sql` — idempotent (`INSERT OR IGNORE`) insert of an official public-PKCE client: no secret, `grant_types` = device-code grant only, `metadata.official: true`, `skip_consent: 1` (exempts it from the stale-client reaper, whose predicate skips `skip_consent` rows).
- **DB-backed validation** — `deviceAuthorization`'s `validateClient` now calls `isDeviceFlowClientAllowed(db, clientId)`: the row must exist, be enabled, and carry `urn:ietf:params:oauth:grant-type:device_code`. Fail-closed on unknown ids, disabled clients, and registered clients without the device grant, on both `/device/code` and `/device/token`. The `UPLOADS_CLI_CLIENT_ID` static allowlist and its special-case comments are removed.
- **Admin panel** — the client now appears in `/admin/oauth` with the official badge, consent/token stats, and a working disable toggle.
- **Docs** — `docs/enrollment.md` covers the managed client, the disable-don't-delete guidance, and deploy ordering.

## Backward compatibility

The client id literal stays `uploads-cli`, so already-configured CLIs are unaffected — nothing changes on the wire, and no CLI-side code changes. Exploration confirmed no existing `oauth_refresh_token`/`oauth_access_token` rows reference the id (the device flow issues Better Auth sessions, not AS tokens), so there is no data backfill. Hosted-MCP dynamic-registration clients are deliberately untouched.

## Deploy note

`pnpm migrate:d1` in `apps/auth` must run before or with the worker deploy (the package's `deploy` script already orders it this way) — the new `validateClient` fails closed if the seeded row is missing.

## Tests

`apps/auth` suite 144/144 green, typecheck clean. New coverage: seeded-row shape assertion, disabled-client rejection, registered-client-without-device-grant rejection, and a reaper test proving a stale seeded row is never a reap candidate.

## Summary by CodeRabbit

* **New Features**
  * Added a managed OAuth client for CLI device-flow authentication.
  * Device authorization now respects the client’s enabled status and supported grant types.
  * Administrators can disable the CLI login client through OAuth client management.

* **Documentation**
  * Added migration and deployment guidance for the managed CLI OAuth client.
  * Documented how to manage and disable the client safely.

* **Bug Fixes**
  * Device authorization now fails safely when the required client registration is missing or invalid.

### DIFF
--- a/apps/auth/migrations/20260719000000_seed_cli_oauth_client.sql
+++ b/apps/auth/migrations/20260719000000_seed_cli_oauth_client.sql
@@ -5,6 +5,13 @@
 -- except: grant_types carries the RFC 8628 device grant, redirect_uris is
 -- empty (device flow has no redirects), and metadata marks it official.
 -- Idempotent: INSERT OR IGNORE keys off the client_id UNIQUE constraint.
+-- skip_consent is seeded as 1: this is a first-party official client, and the
+-- oauth-client-reaper (apps/auth/src/oauth-client-reaper.ts) exempts
+-- skip_consent clients from its stale-client sweep. The device-authorization
+-- plugin never consults skip_consent, so this is semantically a no-op for the
+-- device flow itself but keeps this row from being reaped once it goes stale
+-- (it has user_id NULL and will never accrue oauth_consent/oauth_access_token
+-- rows, since device-flow logins issue Better Auth sessions, not tokens).
 INSERT OR IGNORE INTO oauth_client (
   id, client_id, client_secret, name, redirect_uris, scopes,
   grant_types, response_types, token_endpoint_auth_method, type,
@@ -21,7 +28,7 @@ INSERT OR IGNORE INTO oauth_client (
   '[]',
   'none',
   'web',
-  1, 1, 0, 0,
+  1, 1, 0, 1,
   NULL,
   '{"official":true}',
   CAST(strftime('%s','now') AS INTEGER) * 1000,
--- a/apps/auth/migrations/20260719000000_seed_cli_oauth_client.sql
+++ b/apps/auth/migrations/20260719000000_seed_cli_oauth_client.sql
@@ -1,0 +1,29 @@
+-- Issue #251: the CLI's device-flow client becomes a managed oauth_client
+-- registration instead of a string-allowlisted static id. The client_id stays
+-- 'uploads-cli' so already-configured CLIs keep working. Shape mirrors the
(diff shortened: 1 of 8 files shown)


zachdunn:
Addressed the device.test.ts nitpick in f278c61: verified it was still valid (both negative tests only exercised /device/code, and the sole /device/token test used a never-registered id). Added two tests that claim+approve a device code while the client is valid, then disable it / strip its device-code grant, and assert POST /device/token rejects with 400 `invalid_grant` (the error better-auth's token endpoint actually returns when validateClient fails, vs `invalid_client` on /device/code). apps/auth suite 146/146 green.

PR #260, opened 2026-07-18
feat(auth,api,web): accidental-deletion guards for official clients, last admin, and communal org

zachdunn:
Follow-up to #253: make first-party/system-critical entities hard to delete by accident, in the UI and at the API level. Audit of all destructive admin paths found workspaces already have the gold-standard guard (communal exemption + soft delete + grace/restore); galleries and enrollments have no exposure. This PR closes the gaps that remained.

## Guards added

**Official OAuth clients** — `DELETE /internal/oauth-clients/:clientId` now returns 409 (`official_client`) for clients with `metadata.official: true`; deletion requires deliberately removing the official flag first (PATCH `official: false`). The API proxy relays the 409, and the `/admin/oauth` detail page disables the delete control with a hint (server enforcement is the real guard). Protects the seeded `uploads-cli` client, whose deletion would break CLI login with no automatic re-seed (the migration is journaled).

**Reaper exemption** — the nightly oauth-client reaper now excludes official clients via `json_extract(metadata,'$.official')` in the shared candidate predicate, so the exemption holds in both the listing and the atomic delete statement. Previously a panel-created official client (skipConsent=false, userId NULL) that stayed unused past retention would have been silently hard-deleted.

**Last-admin lockout** — a `hooks.before` guard on the Better Auth admin plugin's user routes blocks: removing or banning yourself, removing/banning the last active admin, and (via `/admin/set-role` and `/admin/update-user`) demoting or banning the last active admin. Without this, one API call could lock every operator out of the admin UI. Handles comma-separated role strings and both string/array role bodies, verified against better-auth 1.6.23's route shapes.

**Communal org** — `DELETE /internal/orgs/:slug` refuses the communal `default` org (403 `protected_org`) even when single-member; previously only multi-member orgs were protected. The slug constant mirrors apps/api's `DEFAULT_WORKSPACE` (noted in-code to keep in lockstep).

## Tests

apps/auth 166/166, apps/api 478/478, typecheck clean across auth/api/web. New coverage: 409 + flag-removal-then-delete flow, proxy 409 relay, reaper official-exemption (stale official survives, non-official reaped), last-admin guard endpoint + helper cases, communal org refusal (with a `default-staging` prefix-confusion negative).

## Summary by CodeRabbit

- **New Features**
  - Protected the final active administrator from removal, banning, or demotion.
  - Prevented deletion of official OAuth clients until they are marked non-official.
  - Official operator-panel OAuth clients are now excluded from automatic cleanup.
  - Prevented deletion of the default communal workspace.

- **Bug Fixes**
  - Admin OAuth deletion now clearly reports official-client conflicts.
  - Official OAuth clients are visibly protected in the admin interface.

- **Documentation**
  - Clarified the safe process for managing and deleting the official CLI OAuth client.

### DIFF
--- a/apps/api/src/routes/admin-ui-oauth-clients.test.ts
+++ b/apps/api/src/routes/admin-ui-oauth-clients.test.ts
@@ -180,6 +180,28 @@ describe("DELETE /admin-ui/oauth-clients/:clientId", () => {
     expect(res.status).toBe(200);
     expect(await res.json()).toEqual({ ok: true });
   });
+
+  it("relays a 409 from the internal endpoint when deleting an official client", async () => {
+    const env = stubEnv(ADMIN_USER, (path, req) => {
+      if (path === "/internal/oauth-clients/c-1") {
+        expect(req.method).toBe("DELETE");
+        return Response.json(
+          {
+            error: "official_client",
+            message: "official clients cannot be deleted; remove the official flag first",
+          },
+          { status: 409 },
+        );
+      }
+      return new Response(null, { status: 404 });
+    });
+    const res = await app().request("/admin-ui/oauth-clients/c-1", { method: "DELETE" }, env);
+    expect(res.status).toBe(409);
+    expect(await res.json()).toEqual({
+      error: "official_client",
+      message: "official clients cannot be deleted; remove the official flag first",
+    });
+  });
 });
 
 describe("AUTH binding outage", () => {
--- a/apps/auth/src/admin-last-guard.test.ts
+++ b/apps/auth/src/admin-last-guard.test.ts
@@ -1,0 +1,217 @@
+/**
+ * Audit guard (accidental-deletion class, see ad736b9's official-client
+ * guard): the stock admin() plugin's `/admin/remove-user` and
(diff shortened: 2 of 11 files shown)


zachdunn:
Addressed the review round in 5d93588:

- **Banned-target guard (valid, fixed):** the last-admin guard now returns early for banned targets — a banned ex-admin isn't an active admin, so removing them can't reduce the active-admin count. The test now asserts that removal succeeds and the active admin survives. The count<=1 branch stays as defense in depth (and remains meaningfully reachable via the set-role/update-user demotion paths).
- **Ban-user coverage (nitpick, added):** sole admin self-ban → 400 with the user not banned; one of two active admins banning the other → 200 with `banned: true` verified in the DB.
- **"Duplicated assertion opening" (skipped, invalid):** there are no duplicated `expect((await res.json())…)` calls at internal-routes.test.ts or internal-oauth-clients.test.ts — the double-paren `(await res.json()) as {...}` cast is the file-wide assertion idiom, one call per test; the suites pass (auth 171/171).
- **CI:** the Lint & Format failure was oxfmt on hand-resolved rebase conflict hunks (rebase commits skip the lint-staged hook) — reformatted.

PR #274, opened 2026-07-19
fix(web,api): folder browsing, responsive layout, people list for the workspace view

zachdunn:
Follow-ups to the settings overhaul (#268), all found while dogfooding the new workspace view.

## Folder browsing was flat

The new files tab listed every key in one flat list: `WorkspaceFileTable`'s client never sent `delimiter=/`, and `GET /me/workspaces/:name/files` only groups keys into folder `prefixes` when a delimiter is given. `listWorkspaceFolder` now always sends it, and the client tests lock the param in. Verified end to end on the local stack: root shows folders, drill-down + breadcrumbs + URL sync work, files appear only at their leaf level.

## Responsive layout

The account shell had only two modes — 3-column desktop and a ≤680px full stack. In between, the fixed 180px nav + 272px rail left the content column narrower than the file grid's minimum, so the table painted over the rail.

- **681–1080px**: new 2-column mode — side nav stays, the rail (usage / details / quick actions) moves below the content with a divider.
- **≤560px**: the file grid drops its size/type columns (via semantic `wft-head__*` classes, not positional nth-child) so filenames keep room instead of ellipsizing to zero width.

## Rail shows the real public URL

`/me/workspaces` previously exposed only a `hasPublicUrl` boolean, so the rail could only say "configured". The endpoint now passes `publicBaseUrl` through (boolean kept for existing consumers — three components branch on it for open-vs-sign logic), and the details rail renders a host-labeled link with a "configured" fallback for an older API.

## People tab lists members

New member-gated `GET /me/workspaces/:name/members`, backed by a shared `membersForOrg` helper in `org-workspaces.ts` (also adopted by the `admin-ui` members proxy, which previously inlined the same fetch). The payload is sanitized to `email` / `name` / `role` / `createdAt` — internal member/user IDs never reach teammates. Behavior change: non-admin members now see the people list instead of a hard "admin access required" error; the invite form stays admin-only.

## Copy

Bracketed labels (`[add]`, `[copy]`, `[invite teammate]`, `[open on github ↗]`, `[MEMBER]`) flattened to the plain lowercase style the rest of the app uses.

## Review pass

A 4-angle cleanup review (reuse / simplification / efficiency / altitude) ran before commit; applied: shared `membersForOrg`, dropping a duplicate per-request org fetch in the members route (`memberWorkspaceOr404` already resolves the org), semantic column classes over nth-child, and two parsing/label simplifications. Noted-not-done: server-side `delimiter` default (would diverge from the flat-by-default contract of the token-scoped `/v1/:workspace/files` twin) and a breakpoint token system (no existing convention; only worth it if a third breakpoint appears).

Tests: 818 pass across web + api (new coverage for the delimiter param, members route sanitization/communal/404, member-row rendering + escaping, rail URL fallback + escaping).

### DIFF
--- a/.claude/launch.json
+++ b/.claude/launch.json
@@ -12,6 +12,15 @@
       "autoPort": true
     },
     {
+      "name": "stack",
+      "runtimeExecutable": "zsh",
+      "runtimeArgs": [
+        "-c",
+        "export NVM_DIR=\"${NVM_DIR:-$HOME/.nvm}\"; [ -s \"$NVM_DIR/nvm.sh\" ] && . \"$NVM_DIR/nvm.sh\" && nvm use --silent; node scripts/dev-stack.mjs"
+      ],
+      "port": 4321
+    },
+    {
       "name": "stack-raw",
       "runtimeExecutable": "zsh",
       "runtimeArgs": [
--- a/apps/api/src/org-workspaces.ts
+++ b/apps/api/src/org-workspaces.ts
@@ -64,6 +64,37 @@ export async function membershipsForUser(env: Env, userId: string): Promise<Memb
   return body as Membership[];
 }
 
+/** A raw member row from the auth worker's `/internal/orgs/:slug/members`. */
+export interface OrgMember {
+  id: string;
+  userId: string;
+  email: string;
+  name: string | null;
+  role: string;
+  createdAt?: string;
+}
+
+/**
+ * Members of an org via `GET /internal/orgs/:slug/members` over the AUTH
+ * binding. Non-ok is an auth-worker outage/bug, surfaced as a 5xx like the
+ * lookups above. Shared by the admin panel (raw rows) and the member-facing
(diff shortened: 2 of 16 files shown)


zachdunn:
### 📎 Attachments

<a href="https://storage.uploads.sh/default/gh/buildinternet/uploads/pull/274/files-desktop.webp"></a>

<a href="https://storage.uploads.sh/default/gh/buildinternet/uploads/pull/274/files-midwidth.webp"></a>

<a href="https://storage.uploads.sh/default/gh/buildinternet/uploads/pull/274/files-mobile.webp"></a>

<a href="https://storage.uploads.sh/default/gh/buildinternet/uploads/pull/274/people-tab.webp"></a>

<sub>Maintained by <a href="https://uploads.sh">uploads.sh</a> — re-uploading a file with the same name updates it everywhere it is embedded.</sub>

PR #310, opened 2026-07-21
feat(api): promotion endpoint for branch-staged attachments

zachdunn:
## Summary

Phase 2a server-side promotion endpoint. `POST /v1/:workspace/github/promote`
copies a workspace's own branch-staged attachments
(`gh/<owner>/<name>/branch/<branch>/<filename>`, written by the CLI staging
lane running in parallel) into the target PR's stable attachment prefix
(`gh/<owner>/<name>/pull/<num>/<filename>`), so the existing managed-comment
gatherer (`github-comment.ts`, which lists that prefix) picks them up
unchanged. Pure workspace-data operation — no GitHub API call, no
installation lookup, no changes to the comment machinery.

## Behavior

- Auth: workspace bearer token, `writeRateLimit`, `requireScope("files:write")`.
- Body: `{ repo: "owner/name", num: number, branch: string }`, validated
  strictly (repo shape incl. dot-only-segment guard, positive int `num`,
  non-empty printable-ASCII `branch`).
- Lists the workspace's own objects under the branch-staged prefix (segments
  sanitized the same way the staged keys are).
- Freshness: a staged file whose `gh.staged-at` D1 tag is older than 30 days
  is skipped (`reason: "stale"`); a missing/unparsable tag is treated as
  fresh, since it's the workspace's own data.
- For each fresh file: reads the object (bytes + R2 custom metadata) and
  writes it to the PR's `pull/<num>/` key via the same `putObject` path every
  other upload goes through — so it inherits the usual budget/usage
  accounting, key-policy checks, and cache-control/`uploaded-at` stamping.
  Content type, visibility, and provenance custom metadata (`client`,
  `client-version`, etc.) are preserved from the source object.
- The copy's D1 metadata is written as a full, self-contained set:
  `gh.repo`, `gh.kind=pull`, `gh.number`, `gh.ref` (`owner/name#num`),
  `gh.branch`, `gh.promoted-at`.
- The staged original is merge-tagged (not replaced, not deleted) with
  `gh.promoted-to` and `gh.promoted-at` — its own `gh.kind=branch` /
  `gh.staged-at` tags survive, so a later reaper phase or a second PR
  re-promoting from the same branch both have what they need.
- Idempotent: re-promoting overwrites the destination copy (last-write-wins,
  same contract as any other overwrite in this API).
- Degrade-safe: a single-file read/write failure is collected into
  `skipped` with a reason instead of failing the whole request.
- Bounded: at most 100 staged files are processed per call; anything past
  that is reported in `skipped` with `reason: "cap_exceeded"` rather than
  silently dropped.
- Response: `{ promoted: string[], skipped: { key, reason }[] }`. An empty
  staging prefix is a 200 with two empty arrays, not an error.

## Out of scope (later phases)

- CLI wiring to call this endpoint (`packages/uploads` — separate lane
  running in parallel; this PR deliberately doesn't touch it).
- Webhook trigger to call promotion automatically on PR open/push.
- The reaper that deletes stale staged originals.

## Test plan

- [x] `pnpm test` (apps/api): all suites pass, including the new
      `github-promote-route.test.ts` (happy path, freshness skip, missing
      `staged-at` treated as fresh, empty prefix, idempotent re-promote,
      metadata correctness on copy AND original, visibility/provenance
      preservation, the 100-file cap, degrade-safe single-file failure,
      auth/scope rejection, validation 400s).
- [x] `pnpm run typecheck` (apps/api, via `wrangler types && tsc --noEmit`).
- [x] `pnpm run lint` / oxfmt — clean on the changed files.

## Summary by CodeRabbit

* **New Features**
  * Added an authenticated endpoint to promote staged branch attachments to a pull request.
  * Promoted attachments retain visibility and metadata, with safeguards for stale or invalid files.
  * Supports repeat promotion without creating duplicate destination objects.
  * Reports successfully promoted and skipped attachments, allowing individual failures without stopping the operation.

* **Bug Fixes**
  * Handles missing freshness metadata safely and prevents stale attachments from being promoted.

### DIFF
--- a/apps/api/src/github-promote.ts
+++ b/apps/api/src/github-promote.ts
@@ -19,8 +19,21 @@ import { storage } from "./storage";
 import { objectVisibility } from "./visibility";
 import type { WorkspaceRecord } from "./workspace";
 
-/** Max staged files processed per call — bounds the work a pathological branch prefix can trigger. */
-export const PROMOTE_STAGED_CAP = 100;
+/**
+ * Max staged files processed per call — bounds the work a pathological branch
+ * prefix can trigger, and keeps the request's subrequest count well inside
+ * the Workers paid-plan ceiling (1000/request; this worker runs on paid).
+ * Each promoted file costs ~9 subrequests: R2 get (download) + R2 head +
+ * R2 put, each inside `putObject`, plus that call's D1 usage-read,
+ * reservation batch, usage-record batch, and metadata-replace batch, plus
+ * this module's own D1 read + write batch to tag the staged original. Fixed
+ * per-request overhead (auth's KV + D1 lookups, the rate limiter, the R2
+ * list page, the batched D1 metadata read) adds ~5 more. At the old cap of
+ * 100 that's ~905 subrequests worst case — over 90% of the ceiling, too
+ * little margin against list pagination or a workspace with extra budget
+ * checks. At 50 it's ~455 (well under half), so this is set to 50.
+ */
+export const PROMOTE_STAGED_CAP = 50;
 
 /** Staged files older than this (by their `gh.staged-at` D1 tag) are skipped, not promoted. */
 const FRESHNESS_WINDOW_MS = 30 * 24 * 60 * 60 * 1000; // 30 days
@@ -145,8 +158,27 @@ export async function promoteBranchAttachments(
           "gh.promoted-at": nowIso,
         },
       });
-      promoted.push(destKey);
+    } catch (err) {
+      // Never leak internal error detail (D1/R2 messages, key policy
+      // internals) to API callers — log the detail server-side and report a
+      // generic reason.
+      console.error(
+        JSON.stringify({
+          message: "promote copy failed",
+          key,
+          destKey,
+          error: err instanceof Error ? err.message : String(err),
+        }),
+      );
+      skipped.push({ key, reason: "copy_failed" });
+      continue;
+    }
 
+    // The copy succeeded — destKey is promoted regardless of what happens
+    // below. Tagging the staged original is best-effort bookkeeping: a
+    // failure here must not un-promote the file or land it in `skipped`.
+    promoted.push(destKey);
+    try {
       // Merge (not replace) onto the staged original: mark it promoted
       // without disturbing its own gh.repo/gh.kind/gh.branch/gh.staged-at tags.
       await setFileMetadata(env.DB, workspaceName, key, {
@@ -154,7 +186,14 @@ export async function promoteBranchAttachments(
         "gh.promoted-at": nowIso,
       });
     } catch (err) {
-      skipped.push({ key, reason: err instanceof Error ? err.message : "copy_failed" });
+      console.error(
+        JSON.stringify({
+          message: "promote: failed to tag staged original",
+          key,
+          destKey,
+          error: err instanceof Error ? err.message : String(err),
+        }),
+      );
     }
   }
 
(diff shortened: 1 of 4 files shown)


zachdunn:
## Review feedback response

Verified each finding against current code and fixed what's still valid. Commit: `fix(api): isolate promote tagging failures from copy error handling`.

### Finding 1 — tagging failure shouldn't un-promote a file — **fixed**

Confirmed: the single `try { putObject(); promoted.push(destKey); setFileMetadata(); } catch { skipped.push(...) }` block meant a `setFileMetadata` failure on the staged original landed the same key in `skipped` even though the copy had already succeeded. Split into two try/catches: the copy (`store.download` + `putObject`) keeps its own try/catch and only pushes to `skipped` on failure; `promoted.push(destKey)` now happens right after the copy succeeds, and the tagging call (`setFileMetadata` on the original) is isolated in its own try/catch that just `console.error`s on failure — it can no longer touch `promoted`/`skipped`. Added a test (`still counts a file as promoted when tagging the staged original fails`) that fails the D1 write scoped to the original key only and asserts `promoted` still contains the destination key and `skipped` is empty.

### Finding 2 — subrequest budget at cap=100 — **fixed (cap lowered to 50)**

Premise correction accepted (paid plan, 1000/request), but did the arithmetic as requested. Per promoted file:

- `store.download(key)` → R2 GET — 1
- inside `putObject`: `existingHead` (R2 HEAD) — 1; `getWorkspaceUsage` (D1 SELECT) — 1; `reserveUploads` (D1 `.batch`, 2 statements, 1 round trip) — 1; `store.upload` (R2 PUT) — 1; `recordUsageSafe`→`applyUsageDelta` (D1 `.batch`) — 1; `replaceFileMetadata` for the copy's `gh.*` tags (D1 `.batch`) — 1
- this module's `setFileMetadata` tagging the original: `getFileMetadata` (D1 SELECT) — 1; write `.batch` — 1

= **9 subrequests/promoted file**.

Fixed per-request overhead: `workspaceAuth` (KV `REGISTRY.get` + the always-paid `findActiveToken` D1 lookup) — 2; `writeRateLimit` (Workers Rate Limiting binding) — 1; the R2 `list()` enumeration page — 1; the batched `getMetadataForKeys` D1 read — 1 = **~5**.

At the old cap of 100: `5 + 9×100 = 905` — over 90% of the 1000 ceiling, too little margin against a second `list()` page or a workspace whose budget check takes an extra path. Lowered `PROMOTE_STAGED_CAP` to 50: `5 + 9×50 = 455`, under half the budget. Documented this arithmetic directly above the constant in `github-promote.ts` so a future reviewer doesn't have to re-derive it. Updated the cap test (`caps at 50 staged files...`, was 100/101 seeded files, now 50/51) — still passes.

### Finding 3 — generic `skipped[].reason`, not the raw error message — **fixed**

Confirmed `reason: err instanceof Error ? err.message : "copy_failed"` was dynamic — a D1/R2 error message would have reached API callers. Copy-failure catch now always reports `reason: "copy_failed"` and logs the real detail server-side via `console.error(JSON.stringify({ message: "promote copy failed", key, destKey, error }))`, matching the logging convention already used in `usage.ts`. Tightened the existing degrade test to assert the exact skip entry (`{ key, reason: "copy_failed" }`) instead of just checking the key is present, so a future regression back to leaking `err.message` would fail the suite.

### Verification

- `pnpm test` (apps/api): 65 files / 728 tests pass (2 new/updated tests for findings 1 and 3; cap test updated for finding 2).
- `pnpm run typecheck` (`wrangler types && tsc --noEmit`): clean.
- `pnpm exec oxlint` / `oxfmt` on changed files: clean.
- Rebased onto `origin/main` after #309 merged (CLI-only changes in `packages/uploads`, no overlap with this PR's `apps/api`-only diff) — rebase was conflict-free, full suite re-verified green after rebase, force-pushed with `--force-with-lease`.

Not merging, not requesting further review per instructions.

PR #311, opened 2026-07-21
feat(cli): auto-promote branch-staged attachments on PR attach

zachdunn:
## Summary

Phase 2b: CLI wiring for promoting branch-staged attachments into a PR once
one exists.

- **Auto-promote on PR attach**: the default `uploads attach` PR path (and
  an explicit `--pr`) now best-effort calls
  `POST /v1/:workspace/github/promote` with the current git branch before
  refreshing the managed comment, so files staged earlier via
  `attach --branch` show up in the same invocation the PR gets its first
  attach. Never for `--issue` targets. `--no-promote` opts out.
- **`uploads attach --promote`**: explicit promote-only mode with zero file
  arguments — resolves the PR the same way the default path does, promotes,
  refreshes the comment, and exits `0` even when nothing was staged. Useful
  right after `gh pr create` when there's nothing new to upload yet.
  `--promote` rejects file arguments (points at the auto-promote default
  instead) and rejects `--branch`/`--issue`/`--no-promote`.
- Branch resolution differs by path on purpose: the default path's
  auto-promote is best-effort (detached HEAD / non-repo silently skips
  promotion, never fails the attach); `--promote` is an explicit action, so
  the same failure surfaces as a `UsageError`.
- `--json` output gets a `promotion` field (`{promoted, skipped}` or `null`
  when the call failed/was skipped); human mode prints a short
  `>> promoted N staged screenshot(s) from branch <b>` note only when
  something actually promoted.
- `ATTACH_HELP` documents the new flags and flow; both `skills/github-screenshots`
  and `skills/uploads-cli` gained a short section on pre-PR staging →
  promotion.
- One changeset (`@buildinternet/uploads` only, minor).

## Contract dependency

Built against #310's contract (`POST /v1/:workspace/github/promote`, body
`{repo, num, branch}`, response `{promoted: string[], skipped: {key,
reason}[]}`), which is **open but not yet merged**. Every promote call is
wrapped so any failure — 404 from an older/self-hosted server that doesn't
have the route yet, 403, network error — degrades to "nothing promoted"
without failing the attach, mirroring how `syncAttachmentsComment` already
falls back to local `gh` when the bot-comment endpoint is unavailable. No
behavior here depends on #310's exact skip-reason strings or edge-case
semantics, only the top-level `{promoted, skipped}` shape.

## Out of scope

- Automatic promotion triggered by GitHub webhooks (e.g. on PR-opened) —
  that's Phase 3.
- Any changes to `apps/api` (owned by the PR #310 branch, currently being
  amended by another agent).

## Test plan

- [x] `pnpm --filter @buildinternet/uploads test` — 579 tests pass,
      including new coverage for: auto-promote payload + ordering before the
      comment sync, never-for-issues, silent skip on 404/network
      error/missing client method/detached HEAD, `--no-promote`, human-mode
      promoted-note (present/absent), `--json` `promotion` field
      (populated/null), `--promote` happy path + empty-staging exit 0,
      explicit `--pr` with `--promote`, still-requires-inferable-PR,
      detached-HEAD UsageError in explicit mode, and every disallowed flag
      combination.
- [x] `pnpm --filter @buildinternet/uploads run typecheck`
- [x] `pnpm --filter @buildinternet/uploads run build`
- [x] `oxlint`/`oxfmt` clean on touched files (pre-existing warnings only,
      untouched files)

## Summary by CodeRabbit

- **New Features**
  - `uploads attach` now automatically promotes branch-staged attachments when attaching to a pull request.
  - Added `--promote` to promote existing staged attachments without uploading files.
  - Added `--no-promote` to disable automatic promotion.
  - Promotion details are included in JSON output and human-readable status messages.

- **Bug Fixes**
  - Promotion failures are handled gracefully so attachments can still be uploaded.

- **Documentation**
  - Added guidance for staging uploads before a pull request exists and promoting them afterward.

### DIFF
--- a/.changeset/attach-promote-branch-staged.md
+++ b/.changeset/attach-promote-branch-staged.md
@@ -1,0 +1,5 @@
+---
+"@buildinternet/uploads": minor
+---
+
+`uploads attach` now auto-promotes branch-staged attachments (`attach --branch`) into a PR's attachment prefix the first time you attach to that PR, before refreshing the managed comment — no extra step needed once a PR opens for a branch you staged files against. Use `uploads attach --promote` (no file arguments) to promote and refresh the comment without uploading anything new, or `--no-promote` to opt out of the automatic behavior. Promotion talks to a new server endpoint and degrades silently (never fails the attach) when that endpoint isn't available yet.
--- a/packages/uploads/src/client.ts
+++ b/packages/uploads/src/client.ts
@@ -232,6 +232,23 @@ export type GithubCommentResult =
       required?: string[];
     };
 
+/** `POST /v1/:workspace/github/promote` request/response (server contract, PR #310). */
+export interface PromoteBranchAttachmentsOptions {
+  repo: string;
+  num: number;
+  branch: string;
+}
+
+export interface PromoteSkip {
+  key: string;
+  reason: string;
+}
+
+export interface PromoteBranchAttachmentsResult {
+  promoted: string[];
+  skipped: PromoteSkip[];
+}
+
 export interface HealthResult {
   ok: boolean;
 }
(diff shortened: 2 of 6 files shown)


zachdunn:
Review nitpick addressed: `promotionNote` now says "staged attachment(s)" instead of "staged screenshot(s)" (attach handles any file type); pluralization/count/branch suffix unchanged. Test updated; 579/579 passing.

Issue #340, opened 2026-07-21
feat(api): uploader attribution (GitHub user) on gh.* uploads


zachdunn:
# In plain terms

Files carry no record of *who* uploaded them — attribution stops at the workspace. With multiple people (or their agents) contributing media to the same repo, we can't show "your in-flight staged media" or "uploaded by @zachdunn" on a PR feed. We want basic attribution to the piloting GitHub user. Explicitly **not a security boundary** — informational only; workspace tokens remain the authz model.

# What it does

- Server-side stamping at upload time (not client-supplied): the auth middleware already resolves the bearer token's D1 row, which carries `minting_user_id` (`apps/api/src/auth-db.ts`). Better Auth's `account` table (shared D1, `providerId='github'`) links that user to a GitHub account id.
- On writes that create `gh.*`-tagged objects (branch staging, `--pr`/issue attaches; arguably all uploads), stamp:
  - `gh.uploader-id` — Better Auth user id (always available when the token is user-minted)
  - `gh.uploader` — GitHub login, resolved from the account link (cache in KV; the `account` row stores the numeric GitHub id, so login resolution may need one GitHub API call or capture at OAuth time)
- Legacy/service tokens with null `minting_user_id`: no tags, degrade silently.
- Consumers: equality filter `?meta.gh.uploader=<login>` gives "your staged media" for free (composes with `gh.status=staged` from the companion issue); `github_pr_activity` feed can surface contributor logins.

# What it is not

- Not authorization: anyone in the workspace can still see/manage all files.
- Not client-asserted: the CLI never sends uploader identity; the server derives it from the token, so it can't be trivially spoofed (though shared tokens dilute it — acceptable for attribution).

# Technical notes

- Decide tag-vs-column: metadata tags reuse the existing EAV store + filters with zero schema work; a real `uploaded_by` column is cleaner but touches more paths. Recommend tags first.
- Capture GitHub login at signup/login time into the user record (or KV) to avoid per-upload GitHub API calls.

PR #347, opened 2026-07-21
feat(mcp): uploader attribution parity for hosted MCP uploads

### LINKED ISSUE #345: feat(mcp): uploader attribution parity for hosted MCP uploads
# In plain terms

Uploader attribution (#340, shipped in #344) stamps `gh.uploader`/`gh.uploader-id` on gh.\*-tagged uploads — but only on the REST path, where `workspaceAuth` resolves the bearer token's `minting_user_id`. Uploads through the hosted MCP server (`agents.uploads.sh`) authenticate with OAuth 2.1 access tokens (JWTs from the uploads-auth AS, #224/#226) and get no attribution, so agent-driven uploads via the hosted MCP are anonymous while identical CLI uploads are attributed. Parity gap, worth closing since agents are the primary users.

# What it does (proposed)

- The OAuth access token already identifies the user (the AS issues JWTs with user + workspace claims). When the MCP worker's auth layer validates the JWT, expose the user id the same way `workspaceAuth` exposes `mintingUserId`.
- Thread it into the shared `putObject` path (or pre-merge tags in the MCP `put`/`files[]` tool the way `routes/files.ts` does): reuse `uploaderTags()` from `apps/api/src/uploader-identity.ts` — same KV cache (`ghlogin:<userId>`), same AUTH internal route, same degrade-to-id-only semantics, same cap guard (`META_MAX_KEYS`).
- Same override rule: server-derived values win over client/tool-supplied `gh.uploader*` pairs.

# What it is not

- Not authorization (same disclaimer as #340) — attribution only.
- The local stdio `uploads mcp` needs nothing: it calls the REST API with the CLI's token, so it inherits REST-path attribution already.

# Notes

- The MCP worker would need the `GITHUB_CACHE` KV + `AUTH` service bindings if it doesn't already have them (check `apps/mcp/wrangler.jsonc`; `pnpm types` after any change).
- Verify which claim carries the user id in the AS-issued JWT (`sub`?) and that it matches the Better Auth user id used by the internal `/users/:id/github-account` route.

zachdunn:
# In plain terms

Uploader attribution (#340, shipped in #344) stamps `gh.uploader`/`gh.uploader-id` on gh.\*-tagged uploads — but only on the REST path. Uploads through the hosted MCP server (`agents.uploads.sh`) authenticate with OAuth 2.1 JWTs instead of `up_` tokens, so agent-driven uploads via the hosted MCP were anonymous while identical CLI uploads were attributed. This closes that gap by threading the same identity through the MCP `put` tool.

# What it does

- The MCP worker's OAuth JWT verification (`src/oauth.ts`) already validates the AS-issued token; `src/index.ts`'s `oauthAuth` now also reads the JWT's `sub` claim and sets it on the same `mintingUserId` context var the REST API uses — verified against the AS's `@better-auth/oauth-provider` source (`sub: user?.id`, no pairwise-subject config on this AS), so it's the same Better Auth user id the internal `/users/:id/github-account` route expects.
- The `put` tool (`apps/mcp/src/tools.ts`) reuses `uploaderTags()`/`hasGithubTags()` from `apps/api/src/uploader-identity.ts` (newly exported as `@uploads/api/uploader-identity`) and applies the same merge rule as the REST PUT hook: gh.\*-tagged metadata gets server-derived `gh.uploader`/`gh.uploader-id` spread in last (so a caller can't spoof those keys), applied once per call so it covers both single-file and multi-file (`files[]`) uploads, and skipped entirely if the merge would exceed the 24-key metadata cap.
- Adds the `GITHUB_CACHE` KV binding and `AUTH` service binding to `apps/mcp/wrangler.jsonc` (same namespace id / service name as `apps/api`), since `uploaderTags()` needs both. The worker's own JWT verification stays a plain JWKS fetch — this binding is only for the internal GitHub-account lookup.

# What it is not

- Not authorization — attribution only, same disclaimer as #340/#344.
- The local stdio `uploads mcp` needed no changes: it calls the REST API with the CLI's token, so it already inherits REST-path attribution.

# Technical notes

- `apps/mcp/src/env.d.ts` gained optional `GITHUB_APP_*` declarations so the shared `uploader-identity.ts` module (which transitively pulls in `apps/api/src/github-app.ts`) typechecks against this worker's `Env`. The MCP worker has no GitHub App configured, so `githubAppConfig()` degrades to `null` there — GitHub login lookups run unauthenticated (lower rate limit), same best-effort-never-throws behavior as every other failure mode in `uploaderTags()`.
- Ran `pnpm types` after the `wrangler.jsonc` change to regenerate `apps/mcp/worker-configuration.d.ts` (gitignored, not committed).

# Test plan

- [x] `pnpm --filter @uploads/mcp test` — 4 new tests in `apps/mcp/test/mcp.test.ts` covering the MCP wiring (not the resolver internals, which have their own suite in `apps/api/test/routes-files.test.ts`): a gh.\*-tagged upload from a JWT user gets `gh.uploader-id` (and `gh.uploader` when the login resolves), a non-gh upload is untouched, the merge respects the 24-key cap, and a JWT with no `sub` claim degrades to no tags.
- [x] `pnpm test` — full suite, 166 files / 2042 tests passing.
- [x] `pnpm typecheck`
- [x] `pnpm check` (oxlint + oxfmt)

Closes #345


zachdunn:
Also rolled in (per Zach): root pnpm script aliases so common per-package tasks don't need the `--filter` incantation — `build:cli`, `test:api`/`test:mcp`/`test:auth`/`test:web`/`test:cli`, `migrate:d1`/`migrate:d1:local` — plus matching AGENTS.md Commands updates.

PR #254, opened 2026-07-18
feat(web): oEmbed for public file and gallery pages

zachdunn:
## In plain terms

Public share pages for files and galleries can now be unfurled by tools that speak oEmbed. Paste a `/f/…` or `/g/…` link into a chat app, notes tool, or other embed consumer and it can resolve media cleanly instead of only seeing a bare URL.

## What it does

- Adds `GET /oembed?url=…&format=json` on the web app (optional `maxwidth` / `maxheight`, CORS-enabled)
- Advertises discovery via `<link rel="alternate" type="application/json+oembed">` on public file and gallery pages (only when the object is fully public)
- Response shapes: images → `photo`, videos → `video` with an inline player snippet, galleries/other → `link` (with cover thumbnail when available)
- Re-resolves through the existing public API so private / missing objects never leak through oEmbed
- Same-origin URL allowlist only (no open-proxy)
- Brief docs: Attach & share, Reference, and `/llms.txt`

## What it is not

- Not an iframe of the HTML page (pages stay `X-Frame-Options: DENY`)
- Not XML oEmbed (`format=xml` → 501)
- Not a change to storage URLs or the API worker

## How to try it

After deploy:

```bash
# discovery on a public file page
curl -sS https://uploads.sh/f/<workspace>/<key> | grep oembed

# resolve oEmbed JSON
curl -sS "https://uploads.sh/oembed?url=https%3A%2F%2Fuploads.sh%2Ff%2F…&format=json"
```

## Technical notes

- Logic lives in `apps/web/src/lib/oembed.ts`; route at `apps/web/src/pages/oembed.ts`
- Reuses `fetchPublicFile` / `fetchPublicGallery` + `isSafeKey`
- Unit coverage in `oembed.test.ts` (parsing, dimensions, photo/video/gallery, auth → 404, HTTP mapping)

## Test plan

- [x] `pnpm --filter @uploads/web test` (158 tests)
- [x] Pre-commit oxlint + oxfmt
- [ ] Manual: load a public `/f/` image page and confirm discovery link + `/oembed` photo response
- [ ] Manual: gallery index returns `link` + thumbnail when a cover image exists

PR #352, opened 2026-07-21
perf: cut D1 rows-read on admin and metadata paths

zachdunn:
## In plain terms

D1 bills primarily on rows read. Admin and account dashboards were doing more work than needed: N org-summary round-trips, counting by loading every member id, over-reading metadata multi-filters, and (on the account shell) reloading the full workspace list just to show one workspace's role and usage.

This keeps the same UI behavior while reading fewer rows and making fewer AUTH/API round-trips.

### Query / index optimizations
- **Admin workspaces list**: one AUTH call to `GET /internal/orgs/summaries` (three aggregate D1 queries) instead of N×3 per workspace
- **Org summary counts**: `COUNT(*)` instead of selecting every member/invite id
- **Metadata search**: multi-filter AND via index-friendly `INTERSECT`
- **Staging reaper**: batch hydrate with `getMetadataForKeys` per workspace
- **Indexes** (new migrations):
  - API: `file_metadata (meta_key, meta_value)`, partial `auth_tokens (minting_user_id)`
  - Auth: `invitation (organization_id, status)`

### New / extended endpoints (dashboard batching)
- **`GET /internal/memberships?userId=&slug=`** — optional single-org membership + `organizationName` on rows
- **Membership authz** — `memberWorkspaceOr404` uses a slug-scoped lookup (one join), not the full list + N org fetches
- **`GET /me/workspaces/:name/summary`** — membership + public URL + usage (workspace rail)
- **`GET /me/workspaces/:name/people`** — members + invites + role (people tab)

Existing list/members/usage endpoints remain for other callers.

## What it is not
- No user-facing CLI package / changeset release
- Not a redesign of multi-workspace orgs (still 1:1 slug === workspace)

## How to try it
After deploy + D1 migrations for api + auth:

1. `/admin` workspaces list still shows member/invite counts
2. Workspace rail still shows role + usage (now via `/summary`)
3. People tab still lists teammates/invites (now via `/people`)
4. Multi-filter metadata search returns the same matches

## Test plan
- [x] Auth internal-routes tests
- [x] API me + admin-ui + file-metadata + staging-reaper tests
- [x] Web api-client tests
- [ ] Apply D1 migrations on staging/prod for new indexes

PR #364, opened 2026-07-21
fix(mcp,docs): carry the canonical vocabulary to the hosted server and docs

zachdunn:
Follow-ups to #361, from auditing the surfaces that PR didn't touch.

## The hosted MCP was still teaching the wrong keys

#361 fixed `METADATA_DESCRIPTION` so the tool schema stops suggesting `page`
and `resolution` — the exact near-miss spellings the CLI now warns about. But
`apps/mcp/src/tools.ts` had its **own hardcoded copy** of that string, so the
hosted server — the one agents actually connect to over the network — kept
telling them to use `page` and `resolution`. The local stdio MCP was fixed;
the hosted one wasn't. It now composes the shared constant, so this can't
drift again.

Its `put` tool also gains `state`/`app`. It deliberately does **not** derive
the EXIF-sourced keys: there's no image decoding in the Workers runtime
(`sharp` isn't available there), so `state`/`app` are the whole canonical
surface on that server. Both are pure pass-through metadata and cost nothing.

This needed four helpers exported from the package's `/mcp` entry point, which
re-exports a curated list rather than everything — hence the patch changeset.

## The docs made a privacy claim that no longer held

The docs said "**EXIF metadata is stripped**." After #361 that's half the
story: EXIF still leaves the uploaded bytes, but `viewport`/`device`/
`software`/`captured` are promoted into metadata that renders publicly on the
file page. Someone reading that sentence would reasonably conclude nothing
from their camera survives.

Reworded to say what actually happens, name the values, state plainly that
they're publicly visible, and give the `--no-auto` opt-out — plus the denials
(GPS, serial numbers, personal-name tags) so the reassurance that *is* true
stays visible.

Docs: EXIF stripped from bytes, allowlist stored as searchable public metadata

## Drive-by

The paragraph immediately below rendered as "Details are in**Reference**" — a
missing space before the link, the Astro whitespace-collapse gotcha. Live on
the public docs since #254. One character.

## Verification

2165 tests pass; typecheck and lint clean. Two new tests pin the hosted MCP
behavior: `state`/`app` reaching stored metadata, and an out-of-enum `state`
being rejected *with the suggestion* — the hosted server now delegates to the
same validator as the CLI, so `state: "post"` says "did you mean after?"
instead of dumping the enum. The docs page was rendered and screenshotted from
a local build (with `uploads screenshot`, dogfooding the tool).

## Checked, nothing needed

`README.md`, `llms.txt`, `sitemap.xml`, and the `reference`/`agents` docs pages
make no metadata or EXIF claims. `uploads find --help` already shows the
canonical example.

### DIFF
--- a/.changeset/hosted-mcp-canonical-metadata.md
+++ b/.changeset/hosted-mcp-canonical-metadata.md
@@ -1,0 +1,8 @@
+---
+"@buildinternet/uploads": patch
+---
+
+Export the canonical metadata helpers (`stateProp`, `appProp`,
+`canonicalMetaFromArgs`, `metadataArgWithCanonical`) from the `/mcp` entry
+point so the hosted MCP server can reuse them instead of keeping its own copy
+of the metadata tool description.
--- a/apps/mcp/src/tools.ts
+++ b/apps/mcp/src/tools.ts
@@ -7,7 +7,10 @@
  */
 import { buildMarkdown, buildScreenshotKey } from "@buildinternet/uploads";
 import {
+  appProp,
   METADATA_DESCRIPTION,
+  metadataArgWithCanonical,
+  stateProp,
   ToolBatchError,
   batchFailureMessage,
   mapBounded,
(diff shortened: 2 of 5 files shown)


zachdunn:
Filed the deferred follow-up from #361 as #365 — surfacing `path`/`state` in the managed 📎 comment.

Worth flagging one thing the draft had wrong: it assumed this was a rendering change. It isn't — `AttachmentItem` never carries metadata, and the server builds it from an R2 `listObjects` with no D1 involved, so it needs a metadata fetch wired into a path that runs on every webhook delivery. Given #352 was explicitly about cutting D1 rows-read, that's worth costing before it ships.

Issue #174, opened 2026-07-15
Consider opt-in/opt-out for overwrite on put/attach


zachdunn:
## Context

Today, uploading to an existing key **always overwrites in place with no confirmation**. That is intentional for stable PR attachment URLs (`attach`, `put --pr`/`--issue`): re-upload the same filename → same public URL → embeds hot-swap.

We now **surface** overwrites after the fact and on **dry-run**:

- Real put: `>> replaced existing object (same URL)` / `"replaced": true`
- Dry-run: `>> would replace existing object (same URL)` / `"replaced": true` without writing

See PR #173.

## Question

Should some paths refuse overwrite unless the caller opts in (or out)?

## Options to evaluate

| Approach | Idea |
| --- | --- |
| **A. Keep default overwrite everywhere** | Status quo; dry-run is the preview |
| **B. Strict only for arbitrary keys** | `--key` / non-`gh/` paths require `--replace` (or fail if exists); `attach`/`--pr` stay hot-swap |
| **C. Global opt-in** | Always require `--force`/`--replace` | Breaks agents |
| **D. Config/env default** | `UPLOADS_NO_OVERWRITE=1` or workspace policy |

## Constraints

- Agents and re-runs must keep a **non-interactive** happy path for PR media
- MCP/stdio tools need the same semantics as the CLI
- Don’t surprise users who re-upload hashed `screenshots/…` keys by accident

## Acceptance ideas (when we pick a design)

- [ ] Documented default for each command path (`attach`, `put --pr`, `put --key`, default put)
- [ ] Flag naming (`--replace` / `--force` / `--no-overwrite`) and JSON error code if refused
- [ ] Tests for refuse vs allow; dry-run still reports would-replace without writing
- [ ] Skill + CLI help updated

## Related

- PR: https://github.com/buildinternet/uploads/pull/173

PR #372, opened 2026-07-22
fix: choose the device-login workspace in the browser, not the terminal (#362)

### LINKED ISSUE #362: Device login: validate/select workspace on the /device approval page, not after
## Observed

Running `uploads login --workspace default --force` when the signed-in user has no membership in `default`:

1. The CLI prints the device code; the user opens `/device`, confirms the code, and the page shows **"Device approved. Return to your terminal — you're signed in."**
2. Back in the terminal, the CLI then fails with `error: no access to this workspace`.

The browser told the user they succeeded; the terminal told them they failed. The workspace-access check runs after device approval (presumably at token mint), so the approval page can't know it's approving a device for a workspace the user can't access.

## Asks

1. **Fail at approval time.** The `/device` page should resolve the requested workspace during approval and show "you don't have access to workspace `<name>`" *instead of* the success panel — the CLI can keep its check as the backstop.
2. **Workspace selection on the `/device` page.** Rather than requiring `--workspace` in advance, let the approval page list the workspaces the signed-in user can access (defaulting sensibly for single-workspace users) and mint for the chosen one. This removes the failure mode entirely and helps first-run UX.

Related: #183 (multi-workspace CLI profiles) — a workspace picker at approval pairs naturally with client-side profiles, but this issue stands alone: even with a single-profile config, approval-time validation and selection fix the confusing success/failure split.

Not pressing — filing so it's on the books (surfaced during the #346 default-workspace migration).

zachdunn:
Closes #362.

## The problem

`uploads login --workspace default`, run by a user with no membership in `default`, told them two contradictory things: the `/device` page showed **"Device approved. Return to your terminal — you're signed in"**, and the terminal then failed with `no access to this workspace`.

The split existed because the workspace was resolved entirely CLI-side *after* the device token exchange. `POST /device/code` carried only `client_id`, so the approval page had no idea which workspace the terminal had asked for and could not validate it.

A second failure came from the same root: an account with two or more workspaces could not run a bare `uploads login` at all — it hard-errored with `multiple workspaces available (a, b); pass --workspace <name>`, demanding a decision from the terminal that the user is better positioned to make in the browser.

## The approach

Workspace selection moves to the `/device` approval page, and the browser becomes authoritative. This follows the pattern `stripe login` uses, which is the right precedent here because — like Stripe's — our credentials are tenant-scoped at mint (`up_<workspace>_…`), so login can't defer tenant choice the way `gh`, `vercel`, or `gcloud` do.

`--workspace` demotes from selector to optional preselect. It isn't removed: `--workspace <name> --create` provisions a *named* workspace, which a picker of existing workspaces structurally cannot express.

### The wire

better-auth's `device_code` row already has a `scope` column that the plugin stores verbatim from `POST /device/code` and echoes back to the client at token exchange. That gives a per-login channel in both directions with no new table:

- **Outbound** — the CLI sends `scope: "workspace:acme"` (plus `create` for `--create`).
- **Inbound** — the page writes the final choice onto the same row; the CLI reads `workspace:<slug>` back out of the token response.

A recorded choice is always written *without* the `create` token, so a surviving `create` is precisely how the CLI distinguishes "the browser decided" from "the browser deferred to me". That convention is load-bearing and is documented in both parsers.

Two alternatives were considered and rejected: a `?workspace=` URL parameter (a crafted link could misrepresent what the terminal requested), and reusing the user-global `oauth_workspace_choice` row (a CLI login would silently re-scope unrelated OAuth/MCP grants and force re-consent).

## What changed

**`apps/auth`** — `deviceWorkspacePlugin`: `GET /api/auth/device/workspace` returns what the terminal requested plus the caller's workspaces; `POST` records the choice, validated against the caller's own memberships. Both are guarded to a row that is pending, unexpired, and unclaimed-or-owned — the same conditions better-auth's own `/device/approve` enforces.

**`apps/web`** — `/device` gains a picker, inline workspace creation for a first-run account, and a **blocking** no-access panel that replaces the approve action entirely. `resolveDeviceWorkspaceState` is extracted as a pure function so the four states are unit-testable. The page moves to a new `devicePageCsp` — `authPageCsp` plus the API origin in `connect-src`, and nothing else (deliberately not `signedInCsp`, which also relaxes `img-src`).

**`packages/uploads`** — sends the scope, reads the browser's decision back, and prefers it over its own flag. A mint 403 now names the workspaces the account *can* reach instead of the bare `no access to this workspace`.

The approve sequence records the choice **before** approving: once the row flips to approved the CLI can claim its token immediately, so a choice written after that would be a race the user experiences as logging into the wrong workspace. Any failure before the approve call aborts without approving.

## Compatibility

Both skew directions are no-ops. An old CLI against this server ignores the echoed scope and mints as it does today; a new CLI against an old server sees an empty scope and falls back to the previous local resolution.

## Testing

`pnpm test` 2193/2193 (174 files) · `pnpm typecheck` exit 0 · `pnpm check` clean.

New coverage: auth endpoint tests against the fake-D1 harness (membership validation, cross-user rejection, expired and already-processed rows, scope round-trip), the four panel states, the CLI precedence across every scope combination, and a drift guard asserting `devicePageCsp` stays identical to `authPageCsp` apart from `connect-src`.

## Not verified live

Browser verification could not be completed: `scripts/dev-stack-common.mjs`'s `seedFixtures` uploads demo fixtures without opting into overwrite, so every boot after the first 409s against the strict-overwrite contract (#174) and the supervisor's group-kill tears down all three already-started workers. Clearing local R2 state fixes exactly one boot, since the preview wrapper's auto-restart immediately triggers a second one. This is pre-existing and unrelated to this branch — filed separately. The page's state machine is covered by static review and typecheck (`.astro` inline scripts have no DOM test coverage in this repo).

## Summary by CodeRabbit

* **New Features**
  * Device login now lets you select or create a workspace in the browser approval page.
  * `uploads login` no longer requires workspace flags; `--workspace` provides an optional preselection.
  * Inaccessible workspace selections are blocked before approval, with clearer recovery guidance.
  * Existing workspace creation workflows using `--workspace <name> --create` remain supported.

* **Documentation**
  * Updated login guidance to explain workspace selection, creation, and token behavior.

### DIFF
--- a/.changeset/device-login-workspace-selection.md
+++ b/.changeset/device-login-workspace-selection.md
@@ -1,0 +1,10 @@
+---
+"@buildinternet/uploads": minor
+---
+
+Device login now picks the workspace in the browser. `uploads login` works with
+no flags for every account — the approval page lists the workspaces you can use,
+creates one if you have none, and refuses to approve a workspace your account
+can't reach instead of reporting success and failing in the terminal.
+`--workspace` becomes an optional preselect; `--workspace <name> --create` still
+provisions by name.
--- a/apps/auth/src/auth.ts
+++ b/apps/auth/src/auth.ts
@@ -20,6 +20,7 @@ import {
   magicLink,
   organization,
 } from "better-auth/plugins";
+import { deviceWorkspacePlugin } from "./device-workspace";
 import { sendAuthEmail } from "./email";
 import { localDemoEnabled, localDemoPlugin } from "./local-demo";
 import * as schema from "./schema";
@@ -548,6 +549,10 @@ function buildAuth(
       // signed-in multi-workspace user record which workspace an OAuth grant
       // should operate on (read back by postLogin.consentReferenceId above).
       workspaceChoicePlugin(db),
+      // Issue #362: GET/POST /device/workspace, letting the /device approval
+      // page resolve and rewrite the workspace a device login mints for
+      // before it approves.
+      deviceWorkspacePlugin(db),
       // Hosted dashboard (`@better-auth/infra`). Omit when the API key is unset.
       ...(dashApiKey ? [dash({ apiKey: dashApiKey })] : []),
       // This endpoint is omitted entirely unless the lifecycle runner supplies
--- a/apps/auth/src/device-workspace.test.ts
+++ b/apps/auth/src/device-workspace.test.ts
@@ -1,0 +1,259 @@
+/**
+ * Issue #362 (auth side): the `/device/workspace` read + write endpoints that
+ * let the approval page resolve and record which workspace a device login
+ * mints for. Driven against the real Better Auth handler via src/index.ts's
+ * `app` on the fake-D1 harness, same pattern as device.test.ts /
+ * workspace-choice.test.ts.
+ */
+import { drizzle } from "drizzle-orm/d1";
+import { eq } from "drizzle-orm";
+import { describe, expect, it } from "vitest";
+import type { AuthEnv } from "./auth";
+import { parseDeviceScope, workspaceScopeValue } from "./device-workspace";
+import { app } from "./index";
+import * as schema from "./schema";
+import { createFakeD1 } from "./test/fake-d1";
+
+function dbEnv(overrides: Partial<AuthEnv> = {}): AuthEnv {
+  return {
+    DB: createFakeD1(),
+    WEB_ORIGIN: "https://uploads.sh",
+    BETTER_AUTH_URL: "https://auth.uploads.sh",
+    ENVIRONMENT: "development",
+    BETTER_AUTH_SECRET_DEV: "test-signing-secret-at-least-32-chars-long",
+    ...overrides,
+  };
+}
+
+/** Seed a user + active session + org memberships; returns the raw bearer token. */
+async function seedSignedInUser(
+  env: AuthEnv,
+  orgSlugs: { slug: string; createdAt: Date }[] = [],
+): Promise<{ userId: string; sessionToken: string }> {
+  const orm = drizzle(env.DB, { schema });
+  const userId = crypto.randomUUID();
+  await orm.insert(schema.user).values({
+    id: userId,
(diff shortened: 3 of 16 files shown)


zachdunn:
Post-merge verification — driven end to end in a real browser against a local three-worker stack, since `.astro` inline scripts have no DOM coverage and this page's state machine was previously backed only by static review.

**Setup:** a real device code requesting `workspace:default`, approved by an account that belongs only to `dev-demo` — the exact scenario from #362.

| Denied — the reported bug | After "Choose a different workspace" |
| --- | --- |
| Device approval page refusing a workspace the account cannot access | Device approval page showing the workspace picker after choosing a different workspace |

**Approve is genuinely unreachable in the denied state**, not merely covered:

- `#confirm` is `hidden`, and `#approve-btn` reports **0 client rects** (a zero box — nothing to click).
- The accessibility tree contains no Approve control at all; the only controls are "Choose a different workspace" and "Deny".
- Server-side the device row stayed `pending`, and the CLI's poll returned `authorization_pending`.

That last point is the fix in one line: the browser says "no access" while the terminal says "still waiting". Before this change the browser said "you're signed in" while the terminal said `no access to this workspace`.

**The round trip works too.** Taking the escape hatch and approving for `dev-demo` made the page report "Device approved — signed in to dev-demo", and the CLI's next poll returned:

```
{"token_type":"Bearer","expires_in":604799,"scope":"workspace:dev-demo"}
```

The terminal receives the *browser's* choice rather than the `workspace:default` it originally asked for — and with no `create` token, so the CLI treats it as authoritative and mints for `dev-demo`.

Console was clean, no CSP violations (the page is on `devicePageCsp`, which adds only the API origin for inline workspace creation).

One note for anyone verifying locally: if the stack dies with a `get-session` 500 during its smoke gate, that's stale local auth D1 — `rm -rf apps/auth/.wrangler/state` and reboot. Unrelated to this change. Fixture re-seeding is fine now thanks to #371.

PR #385, opened 2026-07-22
feat(api): video poster frame foundation

### LINKED ISSUE #299: Video attachments: poster thumbnail linking to /f/ page (inline-looking video in managed comment)
## Problem

External MP4s can't render as an inline video player in a GitHub comment. This is a GitHub platform limitation, not a renderer choice:

- GitHub's comment HTML sanitizer allows `` but **strips `<video>`/`<source>`** tags.
- GitHub's auto-embedded player only triggers for videos hosted on its own `github.com/user-attachments/assets/…` CDN — which a bot/agent can't upload to.

So in the managed `uploads-sh[bot]` comment, a video attachment can only ever be a **link**, never an inline player. In `attachmentsCommentBody` (`apps/api/src/github-comment-render.ts` + `packages/uploads/src/github.ts`), `inferContentType("x.mp4")` returns `video/mp4`, fails the `startsWith("image/")` check, and falls to the bullet-link branch.

## Proposal

Give video attachments an inline **poster/thumbnail image** (a still frame) rendered as an ``, wrapped in a link to the object's `/f/` file page (which already renders a real video player + metadata). That makes a video entry *look* embedded inline while the click-through leads to actual playback.

Sketch:

```html
<a href="{/f/ page url}"><img width="{w}" alt="{name}" src="{poster url}"></a>
```

## Requires

- **Thumbnail generation** for uploaded videos (a still frame / poster). Storage + a way to resolve a video's poster URL. This is the real work; the renderer change is small once a poster URL exists.
- Renderer branch: when a video has a poster URL, emit the `<a></a>` form instead of the bullet link. Mirror in both renderer copies; update the golden fixture.

## Related

Follow-on to the attachment click-through → `/f/` page change (links video bullets to the file page as an interim). This issue is the "make it *look* embedded" step and is intentionally separate because it needs thumbnail generation.

zachdunn:
## What

Poster thumbnails for video attachments (issue #299): generate a still frame
for uploaded videos, store it alongside the source object, and surface it
wherever video attachments already render — the managed GitHub comment (both
copies), the CLI's local golden renderer, and the public `/f/` file page.

The foundation commit (`8dadf80`) already covered generation + storage. This
PR adds the rest of the tasks needed to ship it:

- **Upload wiring** (`apps/api/src/files-core.ts`): `generateAndStorePoster`
  runs on every `PUT /v1/:ws/files/:key` after the object is durably stored.
  It never throws — no poster just means the renderer falls back to its
  pre-#299 bullet link, exactly as before this feature existed.
- **Lifecycle**: posters are carried through delete and visibility changes,
  including the security case — **a private video's poster must never be
  public.** `setObjectVisibility` rewrites the poster's visibility flag
  alongside the source object's; see the `THE security case` comment at
  `apps/api/src/files-core.ts:481`.
- **`_internal/` listing hygiene**: poster objects live under
  `_internal/posters/<key>.jpg`, a reserved namespace excluded from
  `GET /v1/:ws/files` listings so they never appear as user-facing files.
- **Both renderer copies + a new golden**: the managed-comment renderer is
  duplicated between `apps/api` (imports `@uploads/*`) and the CLI package
  (no such imports — the wire type is duplicated intentionally, per the
  brief's self-review). Both now render poster thumbnails, and both are
  pinned by `github-comment-render` (api) / `github-render-golden` (cli)
  against the same fixtures.
- **`gatherAttachments` hydration**: attachments are hydrated with a
  **computed-only** `posterUrl` — never trusted from client-supplied
  metadata, always derived server-side from the object key.
- **`/f/` public file page**: shows the poster with a reserved aspect ratio
  (no layout shift while the video loads), and fixes a metadata leak —
  `apps/api/src/routes/public-files.ts` no longer forwards raw video
  metadata to public clients as generic fields; only the derived
  `posterUrl`/`videoDimensions` cross the trust boundary.
- **Backfill script + runbook** (`scripts/backfill-posters.mjs`,
  `docs/ops.md` § "Video poster thumbnails"): a one-time pass for
  `video/mp4` objects that predate write-time generation, or that came in via
  `POST /sign` (presigned uploads bypass the worker, and therefore
  `generateAndStorePoster`, entirely).

## Rotation spike finding → policy

Task 1's design doc (`.context/2026-07-22-299-video-poster-thumbnails-design.md`)
flagged device-rotated video as the biggest open risk before any code was
written. The spike resolved it against a real iPhone clip (HEVC, 1920x1080,
`rotation=-90`): Cloudflare's Media Transformations binding (`mode: "frame"`)
returned an upright 640x1138 portrait JPEG — rotation is corrected for us.

Resulting policy:
- **Rotation is trusted** — no suppression branch, no follow-up Cloudflare
  Images work needed.
- **No axis swap** — because `MEDIA` already corrects orientation,
  Mediabunny's `getDisplayWidth`/`getDisplayHeight` are already
  rotation-corrected, so `posterImageWidth` must not swap width/height a
  second time. Pinned by a test against a rotated fixture rather than
  assumed.
- **All of `VIDEO_TYPES` allowed**, not just MP4/H.264 — the spike found
  HEVC, QuickTime, and VP9/WebM all transform successfully despite the docs
  specifying MP4/H.264 as the supported case, so the content-type gate
  admits the full set instead of narrowing to MP4 only.

## Kill switches

Three independent switches, documented in `docs/ops.md`:

1. **Flagship flag (preferred — instant, no deploy):**
   ```bash
   wrangler flagship flags update 8371bfe7-9767-4b4d-b75a-37b94d2724f7 \
     video-poster-generation --default off
   ```
   Checked by `posterGenerationAllowed` on every write. **Currently off in
   production** — this feature has not shipped to users yet.

2. **Remove the `MEDIA` binding** from `wrangler.jsonc` and redeploy. Hard
   kill regardless of flag state; survives a Flagship outage; slower (needs a
   deploy).

3. **`POSTER_LIMITER` absence fails closed, not open.** Every other rate
   limiter in this codebase (`WRITE_LIMITER`, `RENDER_LIMITER`) fails *open*
   when its binding is missing. Poster generation is the one exception —
   `posterGenerationAllowed` treats a missing `POSTER_LIMITER` as denied. Not
   a switch to toggle deliberately, but it means a misconfigured or deleted
   binding is safe by default rather than silently permissive.

Any of the three: existing posters keep serving from `_internal/posters/`,
new writes simply stop generating new ones — no data loss, no user-visible
error.

## Ledger / reconcile reasoning

Poster JPEGs are real R2 objects that consume real storage, so their bytes
are counted in the workspace usage ledger (`recordUsageSafe`) exactly like
any other object — both on creation and on the cleanup path when a
replacement can't be postered (the stale frame's bytes are subtracted before
it's deleted). This keeps `reconcileWorkspaceUsage`'s R2-vs-D1 byte count in
agreement; if poster bytes were generated but never ledgered, every video
upload would introduce a small, permanent, and only-growing drift between
the ledger and reality.

## Remaining before flip-on

- **Feature is dark in production.** The Flagship flag defaults off — this
  PR ships infrastructure, not a live user-facing change.
- **Production backfill dry-run is owed.** `scripts/backfill-posters.mjs
  --dry-run` needs to be run against production with a files-scoped token,
  which this session didn't have access to mint. Do this once the flag is on
  and before a real backfill run.
- **A real before/after screenshot of the managed comment is owed, not
  faked.** The brief asked for one, but the feature is dark in production
  and no live poster exists yet to screenshot — generating one would mean
  either flipping the flag on for real (out of scope here) or hand-crafting
  a fixture image and presenting it as a live capture, which would be
  dishonest evidence. Capture this once the flag is flipped on in a
  workspace and a real video has a poster.
- **Flip-on command**, once the above two are done:
  ```bash
  wrangler flagship flags update 8371bfe7-9767-4b4d-b75a-37b94d2724f7 \
    video-poster-generation --default on
  ```

Closes #299.

## Verification

<details>
<summary><code>pnpm typecheck</code> (tail)</summary>

```
apps/web typecheck: Result (110 files):
apps/web typecheck: - 0 errors
apps/web typecheck: - 0 warnings
apps/web typecheck: - 11 hints
apps/web typecheck: Done
apps/mcp typecheck$ wrangler types && tsc --noEmit
apps/mcp typecheck: ✨ Types written to worker-configuration.d.ts
apps/mcp typecheck: Done
```

(Exit 0 across all packages. One real issue was found and fixed during this
verification pass: `apps/mcp`'s program transitively type-checks
`files-core.ts`/`poster.ts` via the `@uploads/api/files` export, but
`apps/mcp`'s `wrangler.jsonc` has no `MEDIA`/`FLAGS`/`POSTER_LIMITER`
bindings. Fixed by making those three optional on the shared `Env`
augmentation in `apps/api/src/env.d.ts` — matching how
`posterGenerationAllowed` already treats them as possibly absent at
runtime — plus a narrowing check before the one direct `env.MEDIA` use in
`files-core.ts` so `apps/api` keeps its non-null guarantee. See commit
`83c39e2`.)

</details>

<details>
<summary><code>pnpm test</code> (tail)</summary>

```
 Test Files  186 passed (186)
      Tests  2319 passed (2319)
   Start at  10:13:11
   Duration  4.93s (transform 7.41s, setup 0ms, import 33.56s, tests 10.15s, environment 9ms)
```

</details>

<details>
<summary><code>pnpm format</code></summary>

```
$ oxfmt
Finished in 1041ms on 546 files using 14 threads.
```

No files needed reformatting beyond the two touched by the typecheck fix
above (already committed).

</details>

<details>
<summary>Golden invariant: <code>pnpm test:api github-comment-render</code> + <code>pnpm test:cli github-render-golden</code></summary>

```
$ pnpm --filter @uploads/api test github-comment-render
 Test Files  1 passed (1)
      Tests  7 passed (7)

$ pnpm --filter @buildinternet/uploads test github-render-golden
 Test Files  1 passed (1)
      Tests  7 passed (7)
```

Both renderer copies pass against the same fixtures.

</details>

## Summary by CodeRabbit

* **New Features**
  * Video uploads now generate poster thumbnails automatically when supported.
  * Video previews display poster images, playback controls, duration, and correct aspect ratios.
  * Public file responses include validated poster URLs and video dimensions.
  * GitHub attachment comments render poster thumbnails with play indicators and video details.
  * Added a tool to generate thumbnails for eligible existing videos.

* **Bug Fixes**
  * Internal poster files and server-only metadata are hidden from user-facing listings and responses.
  * Videos without posters continue to use the existing link-based fallback.


zachdunn:
@coderabbitai review

Issue #506, opened 2026-07-24
Auto-provision a workspace for new users from their GitHub username


zachdunn:
Split out of #479 / #505. Today the answer to "where does a brand-new account land" is the communal workspace. Auto-provisioning a personal workspace at signup would give a better answer — and is probably the prerequisite for #505 (retiring communal on cloud) to be viable at all.

## Proposal

On first signup, derive a workspace slug from the user's GitHub username and provision it, adjusting on conflict.

Signup is already GitHub-gated, so the identity is in hand at the moment of account creation.

## Why it is not a one-liner

The slug rules are stricter than GitHub's username rules, so `username → slug` is a lossy mapping with several cases that must degrade gracefully rather than fail a signup:

- **Case.** `WS_NAME_RE = /^[a-z0-9][a-z0-9-]{1,62}$/` (`apps/api/src/workspace.ts:179`) is lowercase-only; GitHub usernames display mixed case. Lowercase first.
- **Length floor.** That regex requires 2–63 chars. GitHub allows **1-character usernames** (`t`, `k`). Those need a fallback, not a rejection.
- **Reserved names.** `RESERVED_WORKSPACE_NAMES` (`apps/api/src/slug-policy.ts`) includes `api`, `admin`, `new`, `public`, `files`, `me`, `uploads` — all plausible real GitHub usernames.
- **Blocklist false positives.** `validateSlug` runs a profanity blocklist with digit→letter lookalike folding (`0→o`, `1→i`, `3→e`, …) and substring matching. An innocent username can fold into a blocked substring. Signup must never fail — and per that file's existing rule, must never echo *why* — so this needs a silent fallback path.
- **Collision.** GitHub usernames are unique, but the slug namespace is not reserved against them: an unrelated earlier user may already hold `zachdunn`. Needs a documented suffix strategy (`-1`, a short nonce?) and a decision on whether the loser is ever told.

## Decisions needed

1. **Does the auto-created workspace burn a free slot?** `MAX_SELF_SERVE_WORKSPACES = 3` (`packages/billing/src/workspace-cap.ts`). If auto-provisioning counts, every user effectively gets 2 self-chosen workspaces, silently. If it does not, the cap needs a `selfServe`-style carve-out for provisioned-at-signup records.
[...]

Issue #505, opened 2026-07-24
Decide whether the cloud-hosted product should still have a communal workspace


zachdunn:
Split out of #479, where "same workspace" turned out to be the wrong trust boundary for metadata inheritance because the communal workspace (`default`) is a shared tenant every account belongs to.

#479 itself does **not** need this decided — it gates on `isCommunalWorkspace` and moves on. This issue is the standalone product question that surfaced.

## The question

Should cloud keep a single shared tenant that every account is a member of, or should communal become a self-hosted-only concept?

For retiring it on cloud: a shared tenant is a surprising default for a product whose pitch is showing your work on *your* PRs; it forces special cases (reaper exemption, member-cap exemption, reserved slug, account-UI auto-open skip); and it is why cross-tenant concerns like #479's keep needing bespoke gates.

For keeping it: it is the answer to "where do brand-new users land before they have made a workspace," and there is no replacement today. See the auto-provisioning issue filed alongside this — that is probably the prerequisite.

## Scope if we retire it on cloud

Not a config flip. Known surface:

- **`packages/uploads/src/config.ts:46`** — `DEFAULT_WORKSPACE = "default"` is the CLI's default target in a **published npm package**. Every existing install running bare `uploads put` with no `--workspace` targets communal. This is the expensive one: removing it cloud-side breaks those installs until each user reconfigures, and the CLI cannot depend on a private package to learn otherwise.
- **Live public URLs.** Keys are shared-bucket prefixed, so links already pasted into PRs and Slack resolve under the communal prefix. Retiring the workspace is a link-breaking data migration, not a rename.
- **Code sites** (all importing from `@uploads/workspace` after #504): `apps/api/src/slug-policy.ts` reserved name, four `apps/api/src/routes/admin.ts` route defaults, `apps/api/src/routes/internal-billing.ts` member-cap exemption, `apps/api/src/routes/me.ts` `communal` flag, `apps/web/src/pages/console.astro` form default, `packages/email/src/invites.ts` invite framing, plus the workspace reaper exemption.
- **Onboarding**: needs a new landing place for accounts with no workspace.

## Note

If communal survives only on self-hosted, the two deployments diverge in trust model and every cross-tenant feature has to reason about it twice. Worth deciding explicitly rather than by drift. My inclination on #479's specific rule is to gate it in **both** deployments regardless — "these bytes match, inherit the metadata" across unrelated members of a shared tenant is surprising anywhere.

zachdunn:
[...]

Issue #469, opened 2026-07-23
Derived screenshot metadata (path/url/viewport) is lost across the capture-then-attach seam


zachdunn:
## Motivating trace (2026-07-23, PR #468)

An agent followed the recommended capture-early flow for the PR #468 docs screenshot:

1. `uploads screenshot http://localhost:4321/docs/limits --out docs-limits.png --state after` — the hosted copy correctly got the auto-derived metadata: `path=/docs/limits url=… env=local viewport=1520x1400@1x`.
2. After the PR opened: `uploads attach docs-limits.png --pr 468 --state after` — the PR-keyed object (`gh/buildinternet/uploads/pull/468/docs-limits.webp`) got only `gh.*` context and the hand-passed `state=after`. The derived `path` was lost.

The metadata pipeline worked; it just didn't survive the seam between the derivation point (`screenshot`) and the PR-attachment point (`attach`). `path` is one of the highest-value queryable tags (same tier as `state=`), and any agent that captures before the PR exists hits this.

Note the one-command flow already exists (`uploads screenshot <url> --pr N` carries derived meta onto the PR key) — the gap is only the temporal case where the PR doesn't exist yet at capture time.

## Proposed levers, by impact

1. **`screenshot` should participate in branch staging.** Bare `put` on a non-default branch stages branch-keyed and auto-promotes at PR open (issue #403), but `screenshot` writes its own dated `screenshots/<repo>/<date>/…` layout even on a feature branch. If `screenshot` staged branch-keyed like `put` does — carrying its derived `path`/`url`/`viewport`/`env` meta through promotion — the capture-early flow would land full metadata on the PR object with zero agent effort. This is the flow the github-screenshots skill teaches ("stage as you go"), so it closes the seam for the default loop.

2. **Derived metadata rides with content-identical re-uploads.** When `put`/`attach` receives bytes that optimize to an object the workspace already holds, copy the earlier object's derived metadata onto the new key (explicit `--meta`/`--state` flags still win). Alternatively: `--out` writes a sidecar manifest that a later `put`/`attach` of that file reads back. Rescues every two-step workflow, not just this one.

[...]

PR #516, opened 2026-07-24
feat(api): content-identical re-uploads inherit derived metadata

### LINKED ISSUE #479: Content-identical re-uploads inherit the earlier object's derived metadata (server-side)
Follow-up to #469 (lever 2's alternative variant) and PR #473.

## Context

PR #473 closed the capture→attach seam with a client-side sidecar manifest: `screenshot --out file.png` writes `file.png.uploads.json` (derived metadata + SHA-256), and a later `put`/`attach` of that exact file merges it back. That rescues the flow the CLI itself performs, but only when the sidecar survives alongside the file.

## Gap

Any path that loses the sidecar still loses the metadata: the file gets copied/moved without its sidecar, uploaded by a different tool (hosted MCP, a CI step, a second machine), or the sidecar is cleaned up early. The server already has everything needed to close this generally: when an upload's bytes are content-identical to an object the workspace already holds, copy that earlier object's *derived-tier* metadata (path/url/env/viewport — never explicit user metadata, and explicit `--meta`/`--state` on the new upload always wins).

## Sketch

- Requires a content-hash index over workspace objects (or hashing at upload time + a D1 lookup table). Note uploads are optimized (WebP re-encode), so "content-identical" should be evaluated on the *stored* bytes post-optimize, or on a recorded original-bytes hash — needs a design decision.
- Trust boundary: inherit only server-recorded derived metadata (same closed vocabulary the sidecar uses), scoped to the same workspace.
- Interaction with #473: sidecar (explicit local intent) should win over hash inheritance when both apply.

Rescues every two-step workflow rather than only the sidecar-preserving ones, per #469's principle: metadata derivable at capture should survive the lifecycle without the agent re-stating it.

zachdunn:
Closes #479.

## What this does

When an upload's stored bytes match an object the workspace already holds, the earlier object's derived tags (`path`, `url`, `env`, `viewport`, `state`, …) are copied onto the new one.

## Why, given #473 already shipped

#469 closed the capture→attach seam two ways: `screenshot` stages on a branch (#475), and a sidecar manifest carries derived metadata into a later `put`/`attach` of the same bytes (#473). Neither reaches a path where the sidecar cannot exist or does not survive:

- the **hosted MCP** (`agents.uploads.sh`) — no local filesystem, so no sidecar can ever be written;
- a file copied or moved away from its sidecar;
- a CI step, or a second machine, uploading bytes captured elsewhere.

This is a **narrower slice than when #479 was filed** — #475 already covers the dominant capture-then-attach flow. That's worth weighing at review time; see "Cost" below.

## Decisions worth reviewing

**Gated on the communal workspace, not plain same-workspace.** The issue proposed scoping inheritance to the same workspace. That is the wrong boundary: `default` is the one shared tenant every account belongs to, so a byte match there says nothing about two uploads being related. Without the gate, user A's `path=/admin/billing` and `gh.repo=A/private-thing` would land on user B's upload of the same image. Gated in **both** cloud and self-hosted — inheriting across unrelated members of a shared tenant is surprising in either, and a trust rule that varies per deployment is one every future cross-tenant feature has to reason about twice.

Two larger questions this surfaced are split out and do **not** gate this PR: #505 (should cloud keep a communal workspace at all) and #506 (auto-provision a personal workspace from the GitHub username).

**Its own table, not a `file_metadata` row.** The first cut reused `file_metadata_lookup_idx` and shipped no migration. #511 invalidated that: it had just separated the R2 provenance bag (`provenance`) from the user-facing queryable tier (`metadata`), and writing a server-computed `content-sha256` as a tag pushes a 64-char hash straight back into the tier #511 cleaned up — visible in `meta get`, in hydrated listings, in `find --meta`, and burning one of the 24 per-file key slots on every file forever. The migration buys back the clean split, leaves the per-file cap untouched, and removes an ordering hazard (a separate table cannot be wiped by `replaceFileMetadata`, which is delete-then-insert).

**Fail-soft on both index operations.** They run *after* the bytes are durably stored, so an exception would fail a put whose object already exists in R2 — trading a re-upload for metadata the caller could simply re-state. Deliberately the opposite of `getFileMetadata`, which throws because an empty read there would be a lie. This also keeps the feature inert against a database that has not run the migration yet, which matters because the api and its migrations deploy independently.

**A put with no `--meta` cannot use `replaceFileMetadata`.** That path deliberately leaves an existing object's tags untouched, and a replace is delete-then-insert — inheriting through it would wipe tags the caller never asked to change. That branch is additive-only and never overwrites a key the object already carries. This matters because it is the *main* case: a hosted-MCP put with no metadata is exactly what this rescues.

**`gh.*` is not inheritable**, by construction rather than by filter — a repo/PR association is a claim about an upload's context, not a property of its bytes. That is also what stops inherited tags minting phantom PR activity through `recordPrActivityFromMetadata`, which is additionally fed explicit metadata only.

## Refactor carried along

Extracting the merge this needed moved two things into `file-metadata.ts`, beside the caps they enforce: `mergeWithinMetadataCaps` (cap-aware partial merge, explicit wins, overflow dropped rather than thrown — deliberately the same rule as the CLI's `mergeDerivedMeta`, since capture-time derivation and server-side inheritance are one feature to a user) and `metadataByteLength`, now shared with `validateMetadataEntries`.

That closed a real divergence: the validator **exempts server-owned `video.*` keys** from the key-count cap, so a merge that counted them would have silently blocked an inherited key from landing on a video upload. There is a test for it.

`addMissingFileMetadata` is the additive, overflow-dropping counterpart to `setFileMetadata` — one read and one batch, where the first cut read the row, diffed outside, then handed the result to `setFileMetadata`, which read it again.

## Vocabulary duplication

`INHERITABLE_META_KEYS` restates the CLI's `CANONICAL_META_KEYS`. Neither side can import the other: `packages/uploads` is published and carries no `@uploads/*` workspace deps, so it cannot depend on a shared private package, and `apps/api` importing the CLI would invert the dependency and drag CLI deps toward a Worker bundle. This follows the `packages/billing/src/workspace-cap.ts` precedent, with a parity test that reads the CLI source (the vocabulary is not in its public exports, and its `exports` map points at built `dist/`) so drift fails a test rather than silently making a key non-inheritable.

A third option — publishing the vocabulary as its own tiny package — is recorded in the design doc as weighed, not missed. It did not look worth a fourth published package plus release coordination for a 10-entry array, but reviewers may disagree.

## Cost

Every upload now does one extra D1 write (the hash upsert), and every non-communal upload one extra D1 read (the donor lookup). The lookup is issued right after the hash is computed and awaited later, so it overlaps the R2 write rather than stacking on top of it; the upsert stays last on purpose, because writing the hash row after the metadata is what keeps a donor undiscoverable until its tags have landed.

If the verdict is that this is not worth a write on every upload given how much #475 already covers, the honest alternative is closing #479 as covered-enough — that is a legitimate outcome of this review.

## Visible API change

Per #511 a put echoes the tags it stored; inherited keys now appear in that echo, so a client can see what it got without a second round trip.

## Testing

`pnpm test` — 3047 passing (20 new). `pnpm typecheck` — clean. Covers: inheritance on a byte match; never in communal even on an exact match; no cross-workspace; an object is not its own donor on overwrite; vocabulary filtering; deterministic oldest-donor selection; fail-soft when the table is absent; cap overflow dropping rather than failing; the `video.*` exemption; additive merge never overwriting an existing key; and vocabulary parity with the CLI.

No changeset: nothing in `packages/uploads` changed, and a changeset naming an ignored package would block the next npm publish.

## Summary by CodeRabbit

* **New Features**
  * Automatically carries eligible metadata to newly uploaded files with identical content.
  * Preserves existing metadata while applying inherited tags safely within metadata limits.
  * File metadata responses now reflect the metadata actually stored.
  * Metadata inheritance is isolated by workspace and excludes restricted or server-managed fields.

* **Bug Fixes**
  * Prevents metadata overflow from failing uploads by dropping additions that exceed limits.
  * Ensures explicit metadata takes precedence over inherited metadata.

### DIFF
--- a/apps/api/migrations/20260724140000_file_content_hash.sql
+++ b/apps/api/migrations/20260724140000_file_content_hash.sql
@@ -1,0 +1,28 @@
+-- Reverse index from an object's stored-body SHA-256 to the object holding it,
+-- so a content-identical re-upload can inherit the earlier object's derived
+-- metadata (issue #479) on paths where the CLI sidecar cannot survive: the
+-- hosted MCP (no local filesystem), CI steps, and second machines.
+--
+-- Deliberately NOT a `file_metadata` row. #511 established that `metadata`
+-- means the user-facing queryable tag tier (`--meta`, `meta get`,
+-- `find --meta`) while the R2 bag responds as `provenance`. `content-sha256`
+-- is server-computed provenance, so storing it as a tag would push it back
+-- into the tier #511 just separated -- a 64-char hash on every file's visible
+-- tags, burning one of the 24 per-file key slots forever.
+--
+-- The primary key is (workspace, object_key), not the hash: an overwrite of the
+-- same key must update that key's hash rather than accumulate a second row.
+-- Many keys may share one hash, which is the whole point of the lookup index.
+CREATE TABLE file_content_hash (
+  workspace      TEXT NOT NULL,
+  object_key     TEXT NOT NULL,
+  content_sha256 TEXT NOT NULL,
+  updated_at     TEXT NOT NULL,
+  PRIMARY KEY (workspace, object_key)
+);
+
+-- Donor lookup: workspace + hash, ordered oldest-first for a deterministic
+-- winner when several objects share bytes. `updated_at` and `object_key` are in
+-- the index so that ordering is served without a table scan.
+CREATE INDEX file_content_hash_lookup_idx
+  ON file_content_hash (workspace, content_sha256, updated_at, object_key);
--- a/apps/api/src/content-hash.ts
+++ b/apps/api/src/content-hash.ts
@@ -1,0 +1,164 @@
+/**
+ * Content-identical re-uploads inherit the earlier object's derived metadata
+ * (issue #479).
+ *
+ * #469 closed the capture-then-attach seam two ways: `screenshot` now stages on
+ * a branch (#475), and a sidecar manifest carries derived metadata into a later
+ * `put`/`attach` of the same bytes (#473). Neither reaches a path where the
+ * sidecar cannot exist — the hosted MCP has no local filesystem to write one,
+ * and a CI step or second machine never sees it. This closes that remainder
+ * server-side: when an upload's stored bytes match an object the workspace
+ * already holds, the earlier object's derived tags are copied onto the new one.
+ *
+ * Design: `.context/479-content-hash-inheritance.md`.
+ */
+
+import { isCommunalWorkspace } from "@uploads/workspace";
+import { addMissingFileMetadata, getFileMetadata } from "./file-metadata";
+
+/**
+ * The closed set of keys inheritance may copy — a restatement of
+ * `CANONICAL_META_KEYS` from `packages/uploads/src/metadata-vocab.ts`.
+ *
+ * Restated rather than imported, and deliberately so. The vocabulary lives in
+ * `@buildinternet/uploads`, which is **published** and carries no `@uploads/*`
+ * workspace dependencies — extracting the list to a shared private package (the
+ * `@uploads/workspace` pattern from #504) would make the published package
+ * depend on a private one, and importing the CLI package here would invert the
+ * dependency and drag CLI deps toward a Worker bundle. This follows the other
+ * in-repo precedent instead: `packages/billing/src/workspace-cap.ts`, which
+ * restates the shape it needs to keep its package independent.
+ *
+ * Drift is caught by a parity test (`content-hash-vocab-parity.test.ts`) that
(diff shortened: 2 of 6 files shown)


zachdunn:
@coderabbitai review

zachdunn:
Addressed the floating-promise nitpick in `df3102f8`: the catch path around `store.upload` now awaits `inheritedPromise` before rethrowing, so the in-flight donor lookup cannot outlive the request. Behavior is unchanged — it resolves to `{}` rather than rejecting, so this only discards a result — but it matches the no-floating-promises guideline and the two budget releases that catch block already awaits.

1193 api tests passing, typecheck clean.

PR #546, opened 2026-07-28
fix(cli): accept --meta on meta set, and make unknown commands correct themselves

### LINKED ISSUE #545: CLI: metadata DX papercuts for agents (flag inconsistency, unhelpful unknown-command help, error hidden by tail)
Hit while attaching a screenshot to #544 as an agent. Setting metadata on an already-uploaded object took two failed attempts before landing. Each failure is small, but they compound, and all three are agent-ergonomics issues rather than human ones.

## 1. `--meta k=v` on `put`/`list`, but bare `k=v` on `meta set`

```bash
uploads put shot.png --pr 544 --meta path=/admin/metrics   # --meta flag
uploads list --meta path=/admin/metrics                     # --meta flag
uploads meta set <key> path=/admin/metrics                  # bare positional
uploads meta set <key> --meta path=/admin/metrics           # error
```

`meta set` errors with `meta set requires k=v pairs and/or --delete <key>` — a good message, and it correctly writes nothing. But the inconsistency is the trap.

It's actively primed by `put`'s own success tip:

```
tip: add --meta path=/route so this shot is findable by page
```

That tip is correct for `put`, and it's the exact moment an agent learns the metadata concept. When the object is already uploaded (the common case for a follow-up tag), the natural move is to carry `--meta` over to `meta set` — where it fails.

**Suggested fix:** have `meta set` also accept `--meta k=v` as an alias for the positional form. It already parses `k=v` pairs, so this is additive and makes the wrong-but-natural guess just work.

## 2. The unknown-command help dump omits the command you need

```
$ uploads set-metadata
unknown command: set-metadata
...
Essentials:
  put / attach / screenshot / login / whoami / list / delete / doctor / install / update
```

`meta` isn't in that list, so the fallback help can't correct the wrong guess. You have to already know to run `uploads help --all` or `uploads meta --help`.

**Suggested fix:** a "did you mean" for near-misses — `set-metadata`, `set-meta`, `metadata`, `tag` → `uploads meta set`. Cheap and it converts a dead end into a one-step recovery.

## 3. The error prints first, then ~25 lines of help — so `tail` hides it

This is the one I'd prioritize for the agent audience. The full output is:

```
line 1:  unknown command: set-metadata
lines 2-27: banner + usage + Essentials + globals + examples
```

Agents routinely trim command output (`| tail -20`, `| head -30`) to control context. Piping this through `tail` shows the *end* of the help dump and discards the actual error — in my case it surfaced `uploads doctor / install / logout`, which reads like unrelated noise rather than "you used the wrong command name." I only diagnosed it by re-running without the pipe.

**Suggested fix**, any of:
- Send the error to **stderr** and the help to **stdout**, so they can be separated.
- **Repeat the error on the last line** as well as the first.
- Keep unknown-command output **short** — error + "did you mean" + `uploads help --all` pointer — rather than dumping the full command list.

The stderr/stdout split is the most standard and helps humans too.

## Why this is worth fixing

The CLI describes itself as "GitHub screenshot + recording uploads **for agents**", and the repo ships an `agent-dx-cli-scale` rubric and a `cli-for-agents` skill. All three items above are things those docs already argue for — consistent flag surfaces, fast actionable errors, and errors that survive truncation. This is the tool's own guidance applied to itself.

Item 3 in particular has no human-facing symptom, which is probably why it hasn't surfaced before.

zachdunn:
## In plain terms

Tagging a file that is already uploaded took three tries in a real agent
session (#545). The metadata flag is spelled `--meta` on `put`, `attach`,
`screenshot`, and `list`, but `meta set` only took bare `k=v` pairs — so the
obvious carry-over failed. The recovery path failed too: `uploads set-metadata`
answered with a 27-line help dump that never mentions `meta`, and agents that
trim output with `| tail` saw only the tail of that dump, not the error. This
fixes all three.

## What it does

- **`--meta k=v` now works on `meta set`.** Positional pairs still work, and
  both forms can appear in one call. `find` gets the same treatment — it is
  documented as the alias for `list --meta`, so `find --meta k=v` dead-ending
  was the same trap.
- **Unknown commands suggest the right one.** `uploads set-metadata` now
  answers `did you mean: uploads meta set`. Suggestions come from a table of
  spellings agents actually reach for (`set-metadata`, `metadata`, `tag`,
  `upload`, `ls`, `rm`, `capture`, …) plus an edit-distance pass over the
  command catalog, so plain typos (`atach`, `uplaod`) land too.
- **That output is now six lines, not twenty-seven.** Error first, suggestion,
  then pointers to `uploads help --all` and `uploads <cmd> --help`. It fits in
  any reasonable `tail` window, and it skips the network version check the old
  path ran before printing.
- **`--json` gets a real payload.** An unknown command prints
  `{ "error", "code": "USAGE", "didYouMean" }` on stdout instead of nothing.

## What it is not

- Nothing is renamed or removed — every existing spelling still works.
- Suggestions never rewrite the command. The CLI prints the guess and exits 2,
  the same rule the metadata vocabulary already follows for near-miss keys.

## How to try it

```bash
uploads meta set screenshots/myapp/42/shot.png --meta path=/admin/metrics
```

```bash
uploads set-metadata
```

```bash
uploads --json set-metadata
```

## Technical notes

`suggestCommand` lives in `packages/uploads/src/cli-suggest.ts` and scores root
commands from the shared catalog before the alias table, so a real command name
always beats a synonym. Containment (`screenshots` → `screenshot`) applies only
to catalog names; aliases match by distance alone, so a short alias like `tag`
can't swallow unrelated input.

The issue also floated splitting the error onto stderr and the help onto
stdout. Both already went to stderr, and an agent piping `2>&1 | tail` would
still merge them — keeping the output short is what actually survives
truncation, so that is the route taken.

## Test plan

- [x] `pnpm test` — 3434 tests, all passing
- [x] New `test/cli-suggest.test.ts` covers the alias table, typo distance,
      plurals, and the short-output guarantee
- [x] `meta set --meta` and `find --meta` covered in their command tests
- [x] `runCli` tests for the suggestion line, the line budget, and the JSON shape
- [x] Ran the built binary by hand: `set-metadata`, `uplaod`, `xyzzy`, `--json`
- [x] `pnpm typecheck`, `oxlint`, `oxfmt --check` clean on changed files

Closes #545

### DIFF
--- a/.changeset/lucky-moons-tap.md
+++ b/.changeset/lucky-moons-tap.md
@@ -1,0 +1,10 @@
+---
+"@buildinternet/uploads": minor
+---
+
+Show a runnable example when a command is missing an argument. `uploads put`
+now answers `error: put requires at least one file` followed by
+`uploads put ./shot.png --pr 123`, so the fix is copy-pasteable instead of one
+`--help` away. Applies to put, attach, find, delete, meta, gallery, comment,
+screenshot, annotate, and config set, plus the unknown-subcommand errors. With
+`--json` the example comes back as an `example` field on the error payload.
--- a/.changeset/nervous-pugs-invent.md
+++ b/.changeset/nervous-pugs-invent.md
@@ -1,0 +1,11 @@
+---
+"@buildinternet/uploads": minor
+---
+
+Stop hiding the reason a command failed (#545). A missing argument used to
+print the command's whole help block with no error line in it — `uploads put`,
+`find`, `delete`, `meta`, `gallery`, `screenshot`, `annotate`, and
+`config set` all did this. Each now prints one `error:` line naming what is
+missing, plus the `uploads <cmd> --help` hint, so trimmed output still carries
+the reason and `--json` gets a real error payload. An unknown `config`
+subcommand does the same instead of dumping the config help.
--- a/.changeset/olive-hounds-smile.md
+++ b/.changeset/olive-hounds-smile.md
@@ -1,0 +1,11 @@
+---
+"@buildinternet/uploads": minor
+---
+
+Smooth out three metadata papercuts agents hit (#545). `meta set` and `find`
+now accept `--meta k=v`, the same spelling `put`, `attach`, `screenshot`, and
+`list` use, alongside the positional `k=v` form. An unknown command answers
+with a "did you mean" suggestion — `uploads set-metadata` points at
+`uploads meta set` — instead of a help dump that never mentioned `meta`. That
+output is also short now, so an agent that pipes through `tail` still sees the
+error; with `--json` it comes back as `{ error, code, didYouMean }` on stdout.
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -9,6 +9,11 @@ This guide covers the everyday flows. For the full command list, the globals,
 and the exit codes, run `uploads help --all` or see
 [`packages/uploads/README.md`](../packages/uploads/README.md).
 
+Errors are short by design. A failure prints one `error:` line on stderr and a
+hint that names the help you want, so output you trim in a script or an agent
+still carries the reason. A mistyped command suggests the closest real one —
+`uploads set-metadata` answers `did you mean: uploads meta set`.
+
 ## Getting started
 
 ```bash
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -9,9 +9,18 @@ This guide covers the everyday flows. For the full command list, the globals,
 and the exit codes, run `uploads help --all` or see
 [`packages/uploads/README.md`](../packages/uploads/README.md).
 
-Errors are short by design. A failure prints one `error:` line on stderr and a
-hint that names the help you want, so output you trim in a script or an agent
-still carries the reason. A mistyped command suggests the closest real one —
+Errors are short by design. A failure prints one `error:` line on stderr, a
+correct example invocation when an argument is missing, and a hint that names
+the help you want. Output you trim in a script or an agent still carries the
+reason:
+
+```text
+error: put requires at least one file
+  uploads put ./shot.png --pr 123
+hint: uploads put --help
+```
+
+A mistyped command suggests the closest real one —
 `uploads set-metadata` answers `did you mean: uploads meta set`.
 
 ## Getting started
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -216,7 +216,11 @@ uploads attach ./mobile-checkout.png \
 uploads find app=web path=/settings
 uploads meta get screenshots/myapp/42/settings.webp
 uploads meta set screenshots/myapp/42/settings.webp page=onboarding --delete path
+uploads meta set screenshots/myapp/42/settings.webp --meta page=onboarding
 ```
+
+`meta set` and `find` take pairs in either spelling: positional `k=v`, or the
+repeatable `--meta k=v` flag that `put`, `attach`, `screenshot`, and `list` use.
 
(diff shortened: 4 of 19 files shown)


zachdunn:
The new unknown-command output, verbatim:

```
unknown command: set-metadata

did you mean: uploads meta set  Merge-set and/or delete metadata pairs

  uploads help --all    Full command list
  uploads <cmd> --help  Per-command options and examples
```

Old output for the same command was the same first line followed by ~26 lines of banner, essentials list, globals, and examples — with `meta` in none of it.

zachdunn:
Follow-up commit: audited the rest of the CLI for the same masking defect and found a bigger one.

Every "you forgot an argument" path printed the command's **entire help block with no error line in it at all**. Not just truncation-fragile — there was no reason to find, in any window:

```
$ uploads put            # old: ~40 lines of PUT_HELP, exit 2, no "error:" anywhere
$ uploads find           # old: FIND_HELP
$ uploads delete         # old: DELETE_HELP
```

Same for `meta`, `gallery`, `screenshot`, `annotate`, `config set`, and an unknown `config` subcommand (which printed the error *then* dumped CONFIG_HELP after it — the exact tail-masking shape from the issue).

All of them now raise a usage error:

```
$ uploads put
error: put requires at least one file
hint: uploads put --help
```

Two lines, immune to `head` or `tail`, and because it goes through the normal usage-error path, `--json` callers get `{"error": "...", "code": "USAGE"}` on stdout where they previously got nothing.

Docs updated for the error contract, since it's now a promise worth stating: the CLI guide, the package README, and the `uploads-cli` skill each describe short errors, the `did you mean` line, and the JSON shape. The web docs pages don't cover CLI errors or metadata, so nothing to change there.

zachdunn:
Audited this branch against the `cli-for-agents` error guidance and found one line we didn't meet:

> On missing required flags: exit immediately with a clear message and a **correct example invocation**, not a hang.

We had the "clear message" and the fast exit, but of 172 usage errors in the CLI, exactly **two** carried a runnable invocation. Pointing at `--help` is a redirect; the rubric asks for the answer inline.

`UsageError` now takes an optional `example`, which prints under the error line:

```
error: put requires at least one file
  uploads put ./shot.png --pr 123
hint: uploads put --help
```

Applied to the missing-argument and unknown-subcommand families — `put`, `attach`, `find`, `delete`, `meta` (bare, and both `get`/`set` without a key), `gallery`, `comment`, `screenshot`, `annotate`, `config set` — plus the unknown-subcommand errors for `meta`, `gallery`, `github`, and `config`, which now also name the valid set. `--json` carries it as an `example` field on the payload.

Deliberately **not** applied to conflict errors like `--pr and --issue are mutually exclusive`. Those already state the whole rule; an example there is noise.

The rest of the rubric checks out on the current branch:

| Item | Status |
| --- | --- |
| Non-interactive first | Every prompt has a flag; `--non-interactive` errors rather than hanging, and non-TTY stdin is read instead of prompted |
| Layered help | `uploads` → essentials, `uploads help --all`, `uploads <cmd> --help` |
| Examples on every `--help` | Verified programmatically — 0 of the help blocks lack an `Examples:` section |
| stdin / pipelines | `put -`, `--spec -`, `--code-stdin` |
| Fail fast, actionable | Now including the example line (this commit) |
| Idempotency | Stable hash-free keys; re-running `attach` overwrites in place and re-syncs one marker-owned comment |
| Dry-run | `--dry-run` on put, delete, screenshot, update |
| Structured success output | key, url, embedUrl, markdown; `--json` / `--format url\|markdown` |

Issue #480, opened 2026-07-23
Duplicate-comment dedupe never runs while the comment-id cache is warm


zachdunn:
Follow-up to #470 / PR #471, observed during live smoke testing on 2026-07-23.

## Behavior

PR #471's self-healing dedupe (patch the oldest marker comment, delete extras) lives in `findMarkerComment`'s consumer — it only executes when the marker *hunt* runs. `upsertBotComment`'s fast path PATCHes the KV-cached comment id directly with no listing, so while the cache is warm a duplicate marker comment sits untouched.

Verified live: with a duplicate planted on PR #468 and a warm cache, `uploads comment --pr 468` updated the cached comment and left the duplicate. It only collapsed after forcing a cache miss (deleting the cached comment → 404 → cache drop → hunt → dedupe). See the throwaway test on issue #474.

## Consequence

A race-created duplicate (the #470 scenario) can persist for up to the cache lifetime — `COMMENT_ID_TTL` is 30 days, refreshed on each hunt — before the healing pass ever sees it. Eventual healing holds; "heals on the next sync" does not when the cache is warm.

## Proposed

Run the dedupe hunt opportunistically in the cases where cost is acceptable and intent is explicit:

- `uploads comment` (the explicit CLI resync, and its server endpoint when invoked that way): always hunt instead of trusting the cached id. The command is rare and is precisely "make the comment state correct" — one extra listing per invocation is fine.
- Optionally: a low-probability hunt on ordinary syncs (e.g. 1-in-N), or a hunt piggybacked on the `issue_comment` webhook self-heal (#325) which already fires on comment activity.

Keep the fast path for the high-frequency attach/promote syncs — the point is an explicit or cheap trigger that bounds duplicate lifetime to "next `uploads comment`" rather than "next cache miss".

Issue #470, opened 2026-07-23
Managed comment: create race duplicates the bot comment; no re-sync after meta set


zachdunn:
## Observed (PR #468, 2026-07-23)

Two identical `uploads-sh[bot]` 📎 Attachments comments were posted (ids 5061168298 / 5061168332, both created 17:02:25Z), and a later `uploads meta set` on the PR-keyed object did not refresh either comment.

## 1. Create race → duplicate comment

`upsertBotComment` (`apps/api/src/github-comment.ts`) is hunt-then-create with no idempotency guard: KV cached-id → marker hunt → POST when nothing is found. On PR #468 two callers ran concurrently — the GitHub App webhook (PR `opened` at 17:02:24, queued promote + comment) and a CLI `uploads attach --pr 468` moments later. Both hunted while the thread had no marker, both POSTed. Consecutive comment ids in the same second confirm the race.

Aggravating factor: the KV comment-id cache is written after create (last-writer-wins) and the marker hunt stops at the first match, so one duplicate becomes permanently unreachable — every future sync PATCHes the winner and the loser drifts stale forever. The `issue_comment` self-heal (#325-era) doesn't help; it recreates missing comments, it doesn't collapse extras.

**Proposed fix — self-healing dedupe rather than a lock:** when the marker hunt finds more than one marker comment, PATCH the oldest and DELETE the rest. That heals existing duplicates on every subsequent sync and makes the create race retroactively harmless. A CAS-style lock on create would only narrow the window; dedupe closes it.

## 2. `meta set` doesn't refresh the comment

Comment sync runs only from attach/comment/promote (`github-comment-service.ts`) and the webhook. `uploads meta set` writes D1 and stops, so display-relevant metadata added after upload (e.g. backfilling `path=` or changing `state=`) never reaches the rendered comment until some unrelated attach happens. Since #370 renders path/state in the comment, `meta set` on a `gh/…`-keyed object should either trigger the same sync attach does, or at minimum print the established-style stderr hint: `tip: run uploads comment to refresh the PR comment`.

## Repro sketch for (1)

Repo with the App installed; from a branch with staged attachments, `gh pr create` and immediately `uploads attach <file> --pr <new-pr>`. Webhook and CLI race the first comment creation.

PR #554, opened 2026-07-29
fix(comment): collapse a managed attachments comment that lost the create race

### LINKED ISSUE #553: Managed attachments comment created twice (1s apart), leaving a stale orphan
Two managed "📎 Attachments" comments were created on the same PR **one second apart**, leaving a permanent stale orphan. The singleton find-or-create appears not to be atomic against a concurrent path.

Observed on [buildinternet/sunny#764](https://github.com/buildinternet/sunny/pull/764). CLI v0.35.0, GitHub App installed on the repo.

## Timeline

| Time (UTC) | Event |
| --- | --- |
| 14:15:23 | PR #764 opened via `gh pr create` |
| 14:15:34 | comment `5118964984` created — 1 attachment |
| 14:15:35 | comment `5118965075` created — 1 attachment |
| 14:15:47 | comment `5118965075` **edited** → 2 attachments |

Both are `performed_via_github_app.slug = "uploads-sh"`.

The key detail: **both comments were created during my *first* `uploads put`**, ~11s after the PR opened — not one per upload, which is what it superficially looks like. My second `put` 12s later correctly found and edited the newer comment.

## Reproduction

```bash
gh pr create --title … --body-file …            # 14:15:23
uploads put framing.png --pr 764 --alt … --width 760 --meta path=/catalog --state after
uploads put montage.jpg  --pr 764 --alt … --width 860 --meta path=/catalog --state after
```

Nothing was branch-staged beforehand — both files were uploaded with an explicit `--pr` after the PR already existed.

The first `put` returned:

```json
"comment": { "action": "created", "count": 1, "via": "bot" }
```

i.e. the CLI believed it had created exactly one comment.

## Net effect

`5118964984` is orphaned: it still shows only the first attachment, was never updated again, and every subsequent `put` edits `5118965075` instead. So the PR permanently displays two attachment comments, one of them stale and incomplete — visible in the PR today.

## Candidate mechanisms

Two paths both create the singleton and neither dedupes on the `<!-- uploads.sh:attachments ws=default -->` marker:

1. **PR-open promotion racing the explicit `put`.** The `pull_request.opened` webhook (delivered ~11s after open, which lines up) runs promotion + comment creation, while the CLI call arriving at the same moment does its own find-or-create. Both check "does a managed comment exist?", both see none, both create.
2. **The `issue_comment` self-heal reacting to its own comment.** The docs describe the managed comment self-healing on `issue_comment` events. If the CLI creates the comment via the bot and the resulting `issue_comment.created` webhook doesn't recognise it as the App's own managed comment, the handler would "restore" it by creating a second. The 1-second gap fits a webhook round-trip better than it fits two independent user-initiated calls.

(1) feels more likely given the PR-open timing, but the sub-second gap is worth checking against both.

## Suggested fixes

- Make find-or-create atomic per `(repo, pr, workspace)` — an advisory lock or a unique constraint, rather than read-then-create.
- Reconcile on write: if more than one comment carries the marker, keep the oldest, fold its attachments in, and delete the rest. That also repairs PRs already in this state (like #764).
- If the `issue_comment` path is implicated, ignore events whose comment body already carries the workspace's own marker.

Happy to leave #764 as-is if it's useful as a live specimen — say the word and I'll clean it up instead.

zachdunn:
Closes #553.

## Root cause

Two writers, neither atomic. `upsertBotComment` hunts for the marker comment,
misses, and creates — and the `pull_request.opened` promotion webhook and an
explicit `uploads put --pr` can do that in the same window, which is the 1-second
gap in the report.

The second half is why it never healed. Both writers then write the comment id to
the same KV cache key, last-write-wins. Every later sync takes the cached-id fast
path, which never lists, so the marker hunt — and with it the #470 dedupe — never
ran again for 30 days. That's the inverted behavior in the report: the *newer*
comment kept getting edited while the older one went stale, the opposite of the
hunt's oldest-wins rule.

Candidate (2) from the issue is ruled out: `isReconcilableCommentEvent` rejects
`issue_comment.created` outright, so the self-heal never reacts to a fresh comment.

## Fix

**Verify once after a create.** Both racers independently agree the oldest marker
comment wins, fold the current body into it, and delete the rest — including their
own create. No lock needed: oldest-wins is a property of the thread, not of who
got there first, so the racers converge without coordinating. It generalizes to an
N-way race in one pass.

The winner is patched *before* anything is deleted, so a failed fold leaves two
comments for the next hunt rather than deleting the one carrying the current body.
Legacy unnamespaced comments stay adopt-only — a second hit there may belong to
another workspace.

**Cache only a verified id.** The id is written to KV only once verification
confirms our comment is the thread's sole managed one. When it can't — the listing
degraded, or the fold failed — the write is skipped, so the next sync misses the
fast path and hunts, which is exactly the repair. The alternative considered and
dropped was shortening the cache TTL globally; it taxed every workspace's every
sync to hedge a case this already covers.

Reconciliation is create-only by design: a patch can't spawn a new comment, so
there is no new orphan for it to reconcile against.

Mirrored in the local-`gh` fallback path, held at deliberate parity per that
file's contract. It has no id cache, so a degraded verification there heals on the
next sync on its own.

## Cost

One extra listing per *create* — i.e. once per PR/issue coordinate, not per
upload. The patch path (every subsequent sync) is unchanged.

## Tests

14 new tests across both transports: lost race folds and deletes its own; won race
keeps its own and deletes the duplicate; no duplicate → untouched; failed fold
deletes nothing and caches nothing; another workspace's legacy comment never
touched; degraded verification still reports `created` but caches nothing.
Verified they fail against the old code. Full suite 3539 passing, lint and
typecheck clean.

## Not included

sunny#764 is left as-is per the offer in the issue. Its cache entry predates this
change, so `uploads comment --pr 764` (which forces the hunt, #480) will collapse
it on demand once this ships.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved managed attachment comment handling during concurrent updates.
  - Prevented duplicate comments and stale orphan comments from remaining on pull requests.
  - Preserved the oldest managed comment while consolidating newer duplicates.
  - Avoided removing legacy comments belonging to other workspaces.
  - Improved recovery when comment verification or reconciliation encounters an error.

### DIFF
--- a/.changeset/olive-donkeys-shake.md
+++ b/.changeset/olive-donkeys-shake.md
@@ -1,0 +1,9 @@
+---
+"@buildinternet/uploads": patch
+---
+
+Collapse a managed attachments comment that lost a create race. Find-or-create
+was not atomic, so a second writer could create its own comment inside the same
+window and leave a permanent stale orphan on the PR. After creating, the comment
+is now verified once: the oldest marker comment wins, the current body is folded
+into it, and the duplicate is deleted.
--- a/apps/api/src/github-comment.test.ts
+++ b/apps/api/src/github-comment.test.ts
@@ -1327,7 +1327,7 @@ describe("upsertBotComment post-create race reconciliation (issue #553)", () => 
       commentUrl: "https://github.com/acme/web/pull/12#c200",
     });
     expect(deleted).toEqual([]);
-    expect(await kv.get("ghcomment:acme:acme/web#12")).toBe("200");
+    expect(await kv.get("ghcomment:acme:acme/web#12")).toBeNull();
   });
 
   it("never deletes another workspace's legacy comment after a create", async () => {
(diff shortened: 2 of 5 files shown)


zachdunn:
@coderabbitai review

PR #566, opened 2026-07-30
fix(web): width-normalize the compare before layer

### LINKED ISSUE #552: Compare mode: extreme aspect-ratio mismatch leaves blank space beside the before layer
Compare mode: an extreme aspect-ratio mismatch leaves blank space beside the before layer

Parked deliberately during #550 — filing so it's tracked rather than rediscovered.

## Behavior

In Compare mode the **after** image sizes the stage box, and the **before** image is fitted into that box with `object-fit: contain` + `object-position: left top` (`apps/web/src/components/ImagePreview.astro`). That contract is what keeps the two layers registered for realistic pairs: two screenshots of the same page at the same viewport width, differing only in page length, stay width-bound and line up from the top-left corner.

When the two images' aspect ratios differ enough, the before layer becomes **height**-bound instead of width-bound. It then renders narrower than the box, left-anchored, with blank space to its right. Past that point, dragging the divider reveals empty stage rather than image.

Worked example: before 800×3000, after 800×600, 78vh cap ≈ 600px. The host shrink-wraps to 800×600 from the after image; the before layer's contain scale is `min(800/800, 600/3000) = 0.2`, so it renders 160px wide inside an 800px box — roughly 640px of blank space.

## Severity

Low, and it fails in the safe direction: no crop, no distortion, no error. Content is never hidden or misrepresented — there's just dead space. Triggering it takes a several-fold change in page length between the two shots, which is not what a normal before/after pair looks like.

## Why it wasn't fixed in #550

Every plausible fix trades away something better:

- `object-fit: cover` keeps the divider aligned but **crops**, which is wrong for a tool whose whole job is comparing two screenshots.
- Sizing the box from the *larger* of the two images breaks the "after image sizes the box" contract that the rest of the component's sizing now depends on, including the property that the comparison has identical geometry whichever half of the pair you're viewing.

Both are bigger changes than the case warrants. Revisit if real pairs ever hit it.

## If someone picks this up

The relevant rules are the `[slot="first"]` / `[slot="second"]` splits under `[data-mode="fit"]` and `[data-mode="full"]`. Verify against a mismatched pair in **both** modes, and re-check that the layers still share an origin — that took several iterations to get right in #550, and it is easy to regress while changing these rules.

Related trap in the same code, for anyone editing it: the second slot is `width: auto` in fit mode, so its box is 0×0 until the image loads, and Chrome will not lazy-load a 0×0 image. The slot images therefore flip to `loading="eager"` on the first Compare click. Restoring `loading="lazy"` there deadlocks Compare into an empty stage, and neither static review nor a local browser pass with unreachable image URLs will catch it.

zachdunn:
Closes #552.

Compare mode fitted the before layer into the after image's box with `object-fit: contain`, which sizes by whichever axis fits first. On a several-fold page-length mismatch that flips the layer to **height**-bound: it renders a fraction of the box's width, left-anchored, with dead stage beside it.

The issue framed this as a rare, low-severity case, and the dramatic version is. But the same rule was quietly costing every mismatched pair: a 20px length difference put the two layers at slightly different scales, so they never quite registered — the thing Compare exists to do.

## What changed

**Normalize on width, in both modes.** Two screenshots of the same page differ in length, not in width, so width is what puts the same content under the same x in both layers. It's also what other before/after sliders do with mismatched sources. The two mode-specific `[slot="first"]` rules collapse into one:

```css
.image-preview img-comparison-slider img[slot="first"] {
  width: 100%;
  height: auto;
  max-height: none;   /* the fit-mode cap would otherwise squash the ratio */
}
```

**Fit clips, Full width grows.** A before layer longer than the box now overflows it, and the two modes want different answers:

- **Fit** is capped at `--preview-max-h`, so the component's own `.first-overlay` clips the tail. Sizing *this* box to the taller layer would let one long screenshot shrink the whole stage — an 800×3000 before against an 800×600 after would squeeze the after down to a ~187px-wide stamp. That's the trade the issue was right to reject.
- **Full width** has no cap and exists to show the image whole, so the box takes the taller layer's ratio via a `--pair-ratio` custom property measured once both images load. Safe *only* here: the width is the stage's either way, so a long layer can only add height — it can never shrink the other one.

That split also makes the clipping recoverable rather than terminal: the tail Fit hides is one click away on a control that's already there.

The `aspect-ratio: var(--pair-ratio, auto)` fallback means an image that never loads leaves the box exactly as it is today. The "after sizes the box" contract is untouched in Fit, where it's load-bearing.

## Verification

CSS-only behavior, so this was measured rather than eyeballed: a harness that inlines the component's `<style>` block verbatim, loads the real `img-comparison-slider@8.0.7` bundle, and uses the real DOM structure — five pairs × both modes.

| pair | old dead strip | new |
| --- | --- | --- |
| 800×3000 / 800×600 (the issue's example) | 488px of a 610px box | 0 |
| 800×620 / 800×600 (ordinary pair) | 20px | 0 |
| 800×300 / 800×600 (before shorter) | 0 | 0 |
| 800×2400 / 800×3000 (both tall, fit cap active) | 0 | 0 |

In all of them the layers now share an origin, share a width, and keep their aspect ratio. `pnpm test:web` 539 passed, `astro check` 0 errors, Prettier and oxlint clean.

Fit and full-width compare stages, before and after the fix

**Known, deliberate:** when the Full-width box grows, the corner `after` label sits at the bottom of the grown box — in the letterbox strip rather than on the after image's own bottom edge. It still reads as marking that half of the stage, and chasing it costs more than it's worth.

Per the issue's warning, the `[slot="first"]` / `[slot="second"]` splits were re-verified in both modes, and the slot images' `loading="eager"` flip is untouched — restoring `lazy` there still deadlocks Compare into an empty stage.

### DIFF
--- a/apps/web/src/components/ImagePreview.astro
+++ b/apps/web/src/components/ImagePreview.astro
@@ -138,6 +138,29 @@ const counterpartUrl = compare?.src ?? null;
       });
     }
 
+    /**
+     * Full width sizes the compare box to the taller of the two layers, so a
+     * before layer longer than the after image isn't clipped in the one mode
+     * whose whole job is showing the image whole (#552). Only the ratio needs
+     * measuring — the box's width there is already the stage's — so this is a
+     * single custom property, and `aspect-ratio`'s `auto` fallback means an
+     * image that never loads leaves the box exactly as it is today.
+     *
+     * Deliberately not applied in fit mode: that box is capped at
+     * `--preview-max-h`, so sizing it to the taller layer would let one long
+     * screenshot shrink the whole stage — an 800×3000 before against an
+     * 800×600 after would squeeze the after down to a ~187px-wide stamp.
+     * Fit clips instead, and full width is one click away.
+     */
+    const sizePairBox = () => {
+      const layers = [...wrap.querySelectorAll<HTMLImageElement>("img[slot]")];
+      if (layers.length !== 2) return;
+      if (!layers.every((el) => el.complete && el.naturalWidth > 0)) return;
+      const ratio = Math.max(...layers.map((el) => el.naturalHeight / el.naturalWidth));
+      if (!Number.isFinite(ratio) || ratio <= 0) return;
+      wrap.style.setProperty("--pair-ratio", `1 / ${ratio}`);
+    };
+
     // `loading="lazy"` keeps the counterpart out of the initial page fetch,
     // but in fit mode the second slot is `width: auto` — its box derives from
     // its own intrinsic size, which is 0×0 until it loads, and the host
@@ -150,8 +173,10 @@ const counterpartUrl = compare?.src ?? null;
     const activateImages = () => {
       if (imagesActivated) return;
       imagesActivated = true;
-      for (const el of wrap.querySelectorAll<HTMLImageElement>("img[loading='lazy']")) {
-        el.loading = "eager";
+      for (const el of wrap.querySelectorAll<HTMLImageElement>("img[slot]")) {
+        if (el.loading === "lazy") el.loading = "eager";
+        if (el.complete) sizePairBox();
+        else el.addEventListener("load", sizePairBox, { once: true });
       }
     };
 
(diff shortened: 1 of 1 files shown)


zachdunn:
Verified end-to-end on the preview deploy (Workers version `25c66050`) against a real `/f/` page with a real paired upload — 800×3000 before / 800×600 after, paired by `path` + `state`, so this exercises the counterpart lookup, the toggle wiring, and the real optimized bytes rather than the harness's synthetic DOM.

| | box | before layer | dead strip | before clipped |
| --- | --- | --- | --- | --- |
| Fit | 800×600 | 800×3000 | **0** | 2400px tail |
| Full width | 878×3293 | 878×3293 | **0** | **0** |

Same origin and same width in both, and `--pair-ratio` resolved to `1 / 3.75`.

Compare mode in fit on a real file page, before layer spanning the full box

One thing the harness could not have caught, worth knowing for anyone touching this later: **the upload optimizer stored the tall image at 640×2400, not 800×3000**, so the two halves of a real pair can have *different natural widths*. Ratio-preserving, so width normalization handles it — both layers still resolve to the box's width and register — but it does mean the "pairs share an intrinsic width" intuition is not something this component can rely on. It doesn't need to.

Issue #297, opened 2026-07-20
Cross-tenant authorization for the bot-comment endpoint (workspace may target any org's repo)


zachdunn:
## Restrict which workspaces may make the bot comment on a repo (cross-tenant authorization)

**Security / hardening.** Surfaced while designing the self-documenting comment (#298), but this is a **live gap in the already-merged bot-comment endpoint** (#290), independent of that feature.

### The gap

`POST /v1/:workspace/github/comment` takes `{repo, num, kind}`. It renders only the **calling workspace's own** data (good — prevents arbitrary-body injection), but nothing checks that the calling workspace is *entitled to post to that repo*:

- `WorkspaceRecord` has no GitHub org/owner binding.
- The route validates only the repo string format.
- `installationForRepo(repo)` resolves whichever org has the App installed and mints **that org's** token.

The uploads.sh App is installed org-wide across every customer org, so our system holds write tokens for all of them.

### Exploit walkthrough

1. Org A has a workspace + token. Org B has the App installed (as any customer does).
2. Org A uploads media under a free-form key `gh/orgB/repo/pull/N/x.png` in its own bucket.
3. Org A calls `POST /v1/{orgA-ws}/github/comment` with `{repo:"orgB/repo", num:N, kind:"pull"}`.
4. Server gathers org A's media under that prefix, finds **org B's** installation, posts as `uploads-sh[bot]` into org B's PR.

Worse: the attachments marker is global and an App token with write can edit **another author's** comment, so org A can **overwrite** org B's legitimate `uploads-sh[bot]` comment (defacement), not just append.

The communal `default` workspace is the widest exposure — anyone using it can target any org with the App installed.

**Severity:** medium. Bounded to the actor's own uploaded images and to orgs that installed the App, but it enables unsolicited bot comments on / overwrites of arbitrary org PRs — spam, defacement, and reputational risk carried by our bot identity.

### Proposed controls (layered)

1. **Baseline — workspace ↔ org binding (structural).** Registration is already GitHub-gated; capture the GitHub org/owner a workspace belongs to and only allow the endpoint to target repos in a bound org (or explicitly linked repos). This alone kills the A→B vector. Decide multi-org handling and how the communal `default` workspace is constrained (probably: cannot target arbitrary org repos).
[...]

PR #582, opened 2026-07-31
feat(auth,api): capture GitHub login at OAuth-link time

### LINKED ISSUE #580: Capture GitHub login at OAuth-link time (unblocks the collaborator-permission tier)
Follow-up to #297 / PR #577. The deferred collaborator-permission tier of the actor-on-PR gate needs the caller's GitHub **login**, but the Better Auth `account` table stores only the numeric GitHub account id. Today the login is recovered lazily via `GET /user/:id` (`resolveUploaderLogin`, `apps/api/src/uploader-identity.ts`) — unauthenticated/IP-rate-limited when the App isn't installed on the tagged repo, and impossible to audit historically.

Capture the login at OAuth-link time instead:

- When a GitHub account is linked (sign-in or account-link flow in `apps/auth`), persist the GitHub login alongside the numeric id. GitHub's OAuth profile payload already includes `login`, so no extra API call should be needed.
- Expose it on the existing internal lookup (`/internal/users/:id/github-account`) so `resolveUploaderLogin` can prefer the stored login and fall back to the current chain for accounts linked before this lands.
- Update the stored login on re-authentication (logins change on rename; last-write-wins is fine).
- Backfill is explicitly out of scope — pre-existing links keep using the lazy chain.

This is data capture only; it does not implement the collaborator-permission tier itself (that stays deferred per #297's closing analysis, now on PR #577).

Every login that links before this lands is one that can't be checked later without a backfill, which is why capture is worth doing ahead of need.

zachdunn:
## Summary

- Captures the GitHub **login** at OAuth-link time in `apps/auth`, unblocking the deferred collaborator-permission tier of #297 (now on PR #577).
- Adds a `github_identity` table, populated as a side effect of the github social provider's `mapProfileToUser` hook — the only point in Better Auth's OAuth flow that sees the raw profile (`profile.login`), and one that runs on **every** completed GitHub OAuth callback (first link and every re-authentication of an already-linked account). That last part matters: Better Auth's own account create/update hooks only ever see the fixed token/id field set it writes internally, never the raw profile, so there's no supported way to bolt a `login` column directly onto `account` with a working write path. Keying the table by the numeric GitHub account id (not user id) also means a write here never races the user/account rows Better Auth creates afterward for a brand-new sign-up.
- Because `mapProfileToUser` fires on every callback (not just first link/create), last-write-wins holds naturally — a GitHub rename overwrites the stored login on the next sign-in.
- `/internal/users/:id/github-account` now returns `githubLogin` alongside the existing `githubAccountId` (left unchanged).
- `apps/api`'s `resolveUploaderLogin` now prefers the stored login over the lazy, IP-rate-limited `GET /user/:id` GitHub API chain, falling back to that chain unchanged for accounts linked before this lands (backfill is explicitly out of scope, per the issue).

## Migration

- `apps/auth/migrations/20260731120000_github_identity.sql` — new `github_identity` table. Applies to prod automatically on merge (D1 migrations CI).

## Test plan

- [x] `pnpm test` (full root suite) — 264 files / 3635 tests passing
- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] New tests: `apps/auth/src/auth.test.ts` (`upsertGithubLogin` insert/last-write-wins/no-op), `apps/auth/src/internal-routes.test.ts` (`GET /internal/users/:id/github-account` with/without a captured login), `apps/api/test/uploader-identity.test.ts` (`resolveUploaderLogin`/`uploaderTags` preferring the stored login vs. falling back to the GitHub API chain for pre-#580 links)

Closes #580

## Summary by CodeRabbit

* **New Features**
  * GitHub account lookups now include the account’s current login name.
  * GitHub login names are captured and updated during sign-in, including account renames.
  * Existing accounts without a stored login continue to use the available fallback lookup.

* **Bug Fixes**
  * Authentication is no longer blocked if saving GitHub identity details fails.


zachdunn:
@coderabbitai review

Issue #583, opened 2026-07-31
Self-serve bring-your-own R2 bucket per workspace


zachdunn:
Productize the operator-only BYO-bucket path into self-serve: a workspace admin pastes bucket-scoped R2 credentials in the web UI, they're verified end-to-end (auth, bucket access, public DNS), sealed at rest, and used for all subsequent I/O — no operator, no deploy.

Full plan: `docs/superpowers/plans/2026-07-31-self-serve-byo-r2-bucket.md` (includes the deferred D1 → PlanetScale Postgres evaluation).

Decided for v1:
- BYO is declared at workspace creation (settings can attach only to an empty workspace); migration of populated workspaces and richer mappings are future versions.
- Gated by a per-workspace `byoBucketEnabled` record flag, off by default; enable for `buildinternet` and `default` at rollout. Plan/billing gating is a later decision (mechanism ships dark).
- HTTP-credential mode only (no per-customer wrangler bindings); endpoint derived from `accountId`, never customer-supplied.
- Upload rate limits stay enforced on BYO; storage budget does not.
- `r2.dev` public URLs are rejected with "not supported right now" copy; custom domain or signed-only.

Phases:
- [ ] Phase 0 — hardening: lifecycle guards for unprefixed buckets, typed storage-seam errors, strict credential sealing, single `enc:v1:` implementation
- [ ] Phase 1 — API: `GET`/`verify`/`PUT`/`DELETE` `/me/workspaces/:name/storage` + verification pipeline + flag/budget policy
- [ ] Phase 2 — web: create-flow BYO option + settings storage panel/wizard
- [ ] Phase 3 — CLI doctor check, docs, admin read-out

zachdunn:
Implmented

PR #584, opened 2026-07-31
refactor: Self-serve BYO R2 bucket: plan + Phase 0 hardening

zachdunn:
Part of #583.

Two things, deliberately consolidated per our fewer-larger-PRs preference:

**1. The plan** — `docs/superpowers/plans/2026-07-31-self-serve-byo-r2-bucket.md`: productizing the operator-only BYO-bucket path into self-serve (creation-time choice, credential verify pipeline with a public-URL/DNS probe, settings wizard), behind an off-by-default `byoBucketEnabled` record flag. Includes the deferred D1 → PlanetScale Postgres evaluation with revisit triggers.

**2. Phase 0 hardening** — the guards that must exist before customer-owned buckets do:

- **Lifecycle guards for unprefixed buckets.** `teardownWorkspace` no longer walks-and-deletes objects on an unprefixed dedicated-bucket record by default (on a customer bucket that would delete their entire bucket); platform state (KV/D1/galleries) still tears down, the skip is surfaced as `objectsSkipped: "dedicated-bucket"`, and `DELETE /admin/workspaces/:name?hard=1&purgeObjects=1` is the explicit operator opt-in. The retention sweep skips unprefixed records (`retentionDays` unsupported on BYO in v1); reconcile flags unprefixed scans in its result. New predicates `isUnprefixedDedicatedBucket` / `hasCustomerCredentials`.
- **Typed errors out of the storage seam.** `storageConfig` now throws `storage_misconfigured` (unknown binding) and `storage_credentials_unreadable` (decrypt failure) as 503 `ServiceUnavailableError`s instead of raw 500s; the MCP gallery path propagates `AppError` codes instead of flattening them.
- **No plaintext credentials, ever.** New `sealCredentialFieldsStrict` throws `secrets_key_unconfigured` when `WORKSPACE_SECRETS_KEY` is unset instead of silently storing plaintext (the future self-serve write path uses this; existing lenient behavior unchanged for migration paths).
- **One `enc:v1:` implementation.** `add-workspace.mjs` now imports the canonical WebCrypto sealing from `src/secrets.ts` (Node 24 type stripping) instead of duplicating it in Node crypto; round-trip covered by tests.
- Pinned the abuse-report attachment write to the platform bucket with a comment so it never gets routed through workspace storage config.

**Behavior notes for review**
- The automated retention sweep never passes `purgeObjects` — an unprefixed workspace finalized by the sweep leaves objects behind until an operator hard-deletes with the flag. Safety over convenience; flagged in case we want binding-only records to auto-purge.
- The guard keys on "unprefixed", not "unprefixed + customer creds" — conservative on purpose, so the `buildinternet` binding-mode bucket is protected too.

**Testing**
- Full root suite: 264 files / 3631 tests pass.
- New coverage: prefixed teardown still purges; unprefixed skips but removes platform state; `?purgeObjects=1` opt-in; sweep skip; typed-error translation through the route boundary in both workers; strict-seal throw/round-trip; script sealing round-trips via `openCredentialFields`.
- Pre-existing `apps/mcp` direct-vitest failures (MCP SDK v2 migration) verified byte-identical before/after via stash; the root runner is green.

## Summary by CodeRabbit

- **New Features**
  - Added safer handling for dedicated storage buckets during workspace deletion and retention cleanup.
  - Added an option to explicitly purge objects from dedicated buckets.
  - Added clearer storage and credential configuration errors with service-unavailable responses.
  - Preserved typed storage errors in gallery operations.

- **Bug Fixes**
  - Prevented accidental deletion of objects from unprefixed dedicated buckets.
  - Improved detection and reporting of storage configuration issues.

- **Documentation**
  - Added planning documentation for self-serve customer-managed storage support.

### DIFF
--- a/apps/api/scripts/add-workspace.mjs
+++ b/apps/api/scripts/add-workspace.mjs
@@ -34,18 +34,10 @@
 import crypto from "node:crypto";
 import { wranglerKvKey } from "./run-timed.mjs";
 import { sharedAgentLimitFields } from "./workspace-limit-defaults.mjs";
-
-/** Match apps/api/src/secrets.ts: enc:v1: + base64url(iv || ct || tag). */
-function sealField(master, plaintext) {
-  if (!master || !plaintext || String(plaintext).startsWith("enc:v1:")) return plaintext;
-  if (master.length < 16) fail("WORKSPACE_SECRETS_KEY must be at least 16 characters");
-  const key = crypto.createHash("sha256").update(master).digest();
-  const iv = crypto.randomBytes(12);
-  const cipher = crypto.createCipheriv("aes-256-gcm", key, iv);
-  const enc = Buffer.concat([cipher.update(String(plaintext), "utf8"), cipher.final()]);
-  const tag = cipher.getAuthTag();
-  return `enc:v1:${Buffer.concat([iv, enc, tag]).toString("base64url")}`;
-}
+// Node's built-in type stripping (this repo's engines floor is >=24) lets a
+// plain .mjs script import a .ts source file directly, so the enc:v1: sealing
+// format has exactly one implementation — apps/api/src/secrets.ts.
+import { sealCredentialFields } from "../src/secrets.ts";
 
 const [name, ...rest] = process.argv.slice(2);
 const opts = {};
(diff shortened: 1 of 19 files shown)


zachdunn:
Addressed the review:

- **admin-workspace-delete.test.ts guard contract** — done in the commit above: the unprefixed-record test now asserts `envBucket.listCalls === 0` (the scan itself is the hazard, not just the deletes) and seeds a gallery + `onDeleteOrg` to prove platform-state teardown (galleries, org, KV record) still completes alongside the skip.
- **reports.ts via `@uploads/storage`** — deliberately not taking this one. Abuse-report attachments are platform-owned data with no workspace context, and this PR's Phase 0 explicitly pins that write to `UPLOADS_DEFAULT` (see the new comment at the call site) so it can never be routed through workspace storage config once customer-owned BYO buckets exist. Wrapping it in a `createStorage()` helper would re-couple it to the exact machinery we're separating it from, for one call site.

PR #609, opened 2026-08-05
chore: upgrade TypeScript to 7.0.2 across the workspace

zachdunn:
## What

Moves 11 of 12 workspace packages to **TypeScript 7.0.2**, the native Go compiler. `apps/web` is the sole holdout, for a reason upstream controls.

Typecheck on `apps/api` drops from ~2.2s to ~1.1s. The wins are modest at this repo's size; the real value is being on the supported line before the ecosystem moves.

## The constraint that shapes this PR

TypeScript 7's npm package **no longer ships the JS compiler API**. Its main export is just `lib/version.cjs`:

```js
exports.version = version;
exports.versionMajorMinor = "7.0";
```

`ts.createProgram`, `ts.sys`, and the language service now live behind `typescript/unstable/*`. Any tool that does `import ts from "typescript"` and calls compiler APIs breaks at runtime. Using `tsc` as a CLI is fully supported — only the JS-API import path is affected.

Two packages used such tools. Each was handled on its merits rather than pinned by default.

### `packages/ui` — upgraded (build tool changed)

Nothing about this package required an old TypeScript; only tsup's `dts: true` did, since it generates declarations through `rollup-plugin-dts`. `tsc` emits declarations directly, so tsup now builds only the JS bundle:

```
build: "tsup" → "tsup && tsc -p tsconfig.build.json --emitDeclarationOnly --outDir dist"
dts:   true   → false
```

This also lifts the package off a stale `5.6.3` pin, two majors behind the rest of the repo.

Declaration output changes from one bundled `index.d.ts` to per-file `.d.ts` (12 files). `astro check` resolves them cleanly — 150 files, 0 errors.

### `apps/web` — stays on `^6.0.3`

`@astrojs/check@0.9.10` (latest) declares a peer of `typescript: "^5.0.0 || ^6.0.0"`. TS7 is excluded upstream, so this is not a version lag a bump can fix.

Workarounds were tested, not assumed:

- **`astro build` works on TS7** (exit 0) — it runs through esbuild/vite and never touches the TS API. Only `astro check` breaks.
- **Aliasing a second TypeScript** fails: pnpm resolves a peer dep from the host package, so `@astrojs/check` would still receive 7. It would require a real `typescript@6` plus an alias whose `tsc` bin collides in `.bin`, invoked by absolute path.

The payoff wouldn't justify that complexity anyway. `tsconfig.worker.json` — web's only direct `tsc` use — covers **4 files**; `astro check` covers **149**. Upgrading would speed up 4 files and break type coverage on 149.

Tracked in #610 — revisit when Astro ships TS7 support; it becomes a one-line change.

## Behavior changes encountered

TypeScript 7 rejects an inferred `rootDir` (`TS5011`), so the declaration emit needs it explicitly. This is the only TS7-specific code change in the PR.

## Incidental fix

`packages/ui/tsup.config.ts` imports `node:fs` but sat outside every tsconfig `include`, so it had no program and type-aware lint on it failed. This predates this PR — the original file fails identically — and surfaced only because the file is touched here.

Fixed by splitting the config: `tsconfig.json` now covers `src` plus `tsup.config.ts` (typecheck + lint), and a new `tsconfig.build.json` carries `rootDir` for the emit. `@types/node` added at `^26.1.0`, matching the six other packages. Net effect: `tsup.config.ts` is now typechecked, where before it wasn't.

## Verification

| Check | Result |
|---|---|
| `pnpm typecheck` (13 projects) | exit 0 |
| `astro check` | 150 files, 0 errors |
| `pnpm test` | 270 files, 3798 tests passed |
| `pnpm build` (web) | exit 0 |
| `pnpm check` (lint + format) | exit 0 |
| CLI build (`@buildinternet/uploads`) | exit 0, valid `.d.ts` |

A sweep for other fallout found none: no first-party `typescript` imports anywhere, all tsconfigs already on `module: ESNext` + `moduleResolution: bundler` with none of the options TS7 removed, no version pinned in CI, and no docs stating a version.

## No changeset

`typescript` is a devDependency and doesn't affect the published CLI's runtime tree.

## Summary by CodeRabbit

* **Chores**
  * Updated TypeScript tooling across the project for improved compatibility and development support.
  * Refined UI package build configuration to produce type declarations more consistently.
  * Improved type-checking coverage for UI build-related files.
  * Updated Node.js type definitions used during UI development.

* **Developer Experience**
  * Enhanced reliability and consistency of development and package build workflows.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -48,7 +48,7 @@
   },
   "devDependencies": {
     "@types/node": "^26.1.0",
-    "typescript": "^6.0.3",
+    "typescript": "^7.0.2",
     "vitest": "^4.1.10",
     "wrangler": "^4.110.0"
   }
--- a/apps/auth/package.json
+++ b/apps/auth/package.json
@@ -25,7 +25,7 @@
   },
   "devDependencies": {
     "@types/node": "^26.1.0",
-    "typescript": "^6.0.3",
+    "typescript": "^7.0.2",
     "vitest": "^4.1.10",
     "wrangler": "^4.110.0"
   }
--- a/apps/mcp/package.json
+++ b/apps/mcp/package.json
@@ -25,7 +25,7 @@
   },
   "devDependencies": {
     "@types/node": "^26.1.0",
-    "typescript": "^6.0.3",
+    "typescript": "^7.0.2",
     "vitest": "^4.1.10",
     "wrangler": "^4.107.0"
   }
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "portless": "^0.15.4",
     "prettier": "^3.9.6",
     "prettier-plugin-astro": "^0.14.1",
-    "typescript": "^6.0.3",
+    "typescript": "^7.0.2",
     "vitest": "^4.1.10"
   },
   "engines": {
--- a/packages/billing/package.json
+++ b/packages/billing/package.json
@@ -11,7 +11,7 @@
     "test": "vitest run"
   },
   "devDependencies": {
-    "typescript": "^6.0.3",
+    "typescript": "^7.0.2",
     "vitest": "^4.1.10"
   }
 }
--- a/packages/comment-config/package.json
+++ b/packages/comment-config/package.json
@@ -15,7 +15,7 @@
   },
   "devDependencies": {
     "@types/node": "^26.1.0",
-    "typescript": "^6.0.3",
+    "typescript": "^7.0.2",
     "vitest": "^4.1.10"
   }
 }
--- a/packages/email/package.json
+++ b/packages/email/package.json
@@ -12,7 +12,7 @@
     "test": "vitest run"
   },
   "devDependencies": {
-    "typescript": "^6.0.3",
+    "typescript": "^7.0.2",
     "vitest": "^4.1.10"
   }
 }
--- a/packages/errors/package.json
+++ b/packages/errors/package.json
@@ -12,7 +12,7 @@
     "test": "vitest run"
   },
   "devDependencies": {
-    "typescript": "^6.0.3",
+    "typescript": "^7.0.2",
     "vitest": "^4.1.10"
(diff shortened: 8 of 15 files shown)


zachdunn:
@coderabbitai review

zachdunn:
Declining this one — I think the convention is being applied outside the scope it was written for.

The `lib: ["ES2022"]` rule in AGENTS.md is stated with its rationale attached:

> TypeScript strict, ESM only, `lib: ["ES2022"]` (no DOM — the Workers types own globals like `crypto.subtle.timingSafeEqual`).

That reasoning is about **Workers runtime code**, where the Workers types own the globals and pulling in `lib.dom` would shadow them. `packages/ui` is the opposite case: a React component library that renders in the browser and ships `DOM`/`DOM.Iterable` deliberately. `apps/api`, `apps/auth`, and `apps/mcp` — the actual Workers — all correctly use `lib: ["ES2022"]`.

I did test the suggestion rather than dismiss it. `tsc --noEmit` with `lib: ["ES2022"]` reports **0 errors** today, so it looks safe on the surface. But that's incidental: `skipLibCheck: true` suppresses `@types/react`'s own dependence on `lib.dom`, and no source file currently happens to reference a DOM global directly (the only `document` hits in this package are inside CSS comments).

It breaks the moment anyone writes ordinary UI-library code. Adding a single typical function to `src/`:

```ts
export function focusIt(el: HTMLDivElement | null) {
  el?.focus();
}
```

```
error TS2812: Property 'focus' does not exist on type 'HTMLDivElement'.
Try changing the 'lib' compiler option to include 'dom'.
```

A DOM ref is routine in a component library, so this would be a latent trap rather than a hardening.

Worth noting on scope too: this PR is a TypeScript version upgrade and didn't introduce the `lib` setting. Line 18 only appears in the diff because `tsup.config.ts` was added to `include` so the file finally lands in a program and can be typechecked.

Issue #613, opened 2026-08-10
Consolidate the three API surfaces onto one resource-rooted, auth-orthogonal path scheme


zachdunn:
## Problem

The API currently exposes **three parallel surfaces for the same resources**, keyed by how the caller authenticates:

- `/me/workspaces/:name/<resource>` — session cookie (web app), ~30 routes in `apps/api/src/routes/me.ts`
- `/v1/:workspace/<resource>` — bearer token (CLI / MCP)
- `/public/...` — anonymous

The caller's auth method shapes the URL. Every workspace resource exists at two (sometimes three) paths with slightly different response shapes, and every new feature has to pick a surface or ship on both. `/me/workspaces/:name` as the de facto base of the session API is the visible symptom.

There's also a latent routing smell: the bare `/v1/:workspace` wildcard sits at the same path level as `/v1/tokens`, `/v1/reports`, `/v1/render`, etc., and only works via reserved-name luck.

## End goal

One canonical, resource-rooted surface; authentication orthogonal to the URL (the GitHub model — `/repos/:owner/:repo/issues` is the same path for every credential type):

- **`/v1/workspaces/:workspace/<resource>` becomes the single canonical surface**, accepting either a session cookie or a bearer token. Middleware resolves "who are you + can you touch this workspace" per credential type; handlers don't care.
- **`/me` shrinks to genuine viewer identity**: `/me` (profile), `/me/workspaces` (memberships). Nothing resource-shaped.
- **`/v1/:workspace/...` (bare wildcard) retires**, eliminating the reserved-name collision class. `/v1/workspaces` is already mounted (self-serve registration), so the namespace exists.
- **`/public` stays** — anonymous access is genuinely a different contract (never signs URLs) and keeping it at a separate root is honest.

Prior art considered: GitHub (recommended model), Cloudflare (`/accounts/:id/...`, validates explicit path scoping), Stripe (header-scoped — rejected: web sessions span multiple workspaces, and opaque URLs hurt debuggability), Vercel (`?teamId=` — anti-model).

## Warts to fix in passing

Routes get touched anyway, so:

- `file-url` living outside `files/`
- `PATCH /files/visibility` and `DELETE /files` keying off query params
- five separate `github` sub-routers mounted at the same `/v1/:workspace/github` prefix

## Migration (no flag day)

1. Stand up dual-auth middleware + canonical routes; old paths become cheap Hono aliases (rewrite-and-forward to the canonical handlers).
[...]

PR #615, opened 2026-08-10
refactor(api): dual-auth canonical files vertical at /v1/workspaces/:workspace (#613 phase 1)

zachdunn:
Phase 1 of #613: one canonical, resource-rooted API surface with authentication orthogonal to the URL. This phase establishes the pattern with the **files vertical**; later phases migrate the remaining verticals mechanically.

## What's here

- **`dualWorkspaceAuth()` middleware** (`apps/api/src/dual-workspace-auth.ts`): a bearer token delegates to the existing `workspaceAuth` (hash match, D1 scopes); otherwise the session cookie resolves org membership exactly like `me.ts`'s `memberWorkspaceOr404` (uniform 404, no existence-probing). Either path sets identical `WorkspaceVars`; session callers get every `FILE_SCOPES` since membership is all-or-nothing on the file plane — handlers never learn which credential arrived.
- **Canonical files vertical** at `/v1/workspaces/:workspace/files` (`apps/api/src/routes/workspace-files.ts`): list, search, facets, `file-url` (now under `files/`, per the issue's wart list), visibility PATCH, delete (path-keyed — the query-param wart fixed for the canonical path).
- **Aliases, not breaks**: the old `/me/workspaces/:name/files*` handlers are replaced with a rewrite-and-forward to the canonical handlers; `/v1/:workspace/files*` keeps working untouched. Response shapes on old paths are unchanged — the pre-existing route tests pass unmodified.
- **16 new tests** covering both credential types on the canonical paths, wrong-workspace membership rejection, and token-workspace mismatch.

## Verification

- `apps/api`: 125 test files, 1556 tests green (1540 pre-existing + 16 new); `tsc --noEmit` clean.

## Shape divergences surfaced (design work for later phases, documented in `.context/613-api-consolidation-plan.md`)

- List: session `{files, prefixes, cursor}` (folder-aware, metadata hydrated) vs bearer `{items, cursor, prefixes}` (opt-in metadata + folded search). Canonical adopts the session shape; reconciling the bearer path is flagged as real design work, deliberately not attempted here.
- Visibility PATCH keeps query-param keying even on the canonical path: Hono's `:key{.+}` catch-all doesn't reliably match a static suffix, so this wart isn't cheap — punted with a note.

## Coordination

#614 adds `GET /me/workspaces/:name/files/by-path` on the old convention. Whichever merges second folds it in — with this phase's alias pattern that's one canonical route + one alias line (called out in the plan doc).

## Summary by CodeRabbit

* **New Features**
  * Added workspace file operations accessible through either session authentication or bearer tokens.
  * Added folder-aware file listing, search, facets, URL resolution, visibility updates, and deletion.
  * Added consistent access controls, validation, rate limiting, and error responses.
  * Preserved existing file-management routes and aliases.

* **Tests**
  * Added coverage for authenticated, unauthenticated, invalid-token, and non-member access scenarios.

### DIFF
--- a/apps/api/src/dual-workspace-auth.ts
+++ b/apps/api/src/dual-workspace-auth.ts
@@ -1,0 +1,82 @@
+/**
+ * Dual-auth workspace guard for the canonical `/v1/workspaces/:workspace/*`
+ * surface (issue #613 phase 1). Resolves "who is this + can they touch this
+ * workspace" from EITHER credential type and sets the same `WorkspaceVars`
+ * either way, so downstream handlers — including `requireScope` — never need
+ * to know which one arrived. See `routes/workspace-files.ts` for the first
+ * vertical built on this.
+ */
+import { NotFoundError, UnauthorizedError } from "@uploads/errors";
+import type { MiddlewareHandler } from "hono";
+import { FILE_SCOPES } from "./auth-db";
+import { membershipsForUser, workspacesFromMembership } from "./org-workspaces";
+import {
+  requireSessionUser,
+  sessionAuth,
+  type SessionUser,
+  type SessionVars,
+} from "./session-auth";
+import { loadWorkspaceRecord, workspaceAuth, type WorkspaceVars } from "./workspace";
+
+/** Combined context vars for routes reachable by either auth path. */
+export type DualAuthVars = {
+  Variables: WorkspaceVars["Variables"] & SessionVars["Variables"];
+  Bindings: Env;
+};
+
+/**
+ * Caller's membership for `name`, or a uniform 404 (never a 403 — no
+ * existence probe). Deliberately a small standalone duplicate of
+ * `memberWorkspaceOr404` in `routes/me.ts` rather than an import: this file
+ * needs to stay free of any `routes/me.ts` dependency so `routes/me.ts` can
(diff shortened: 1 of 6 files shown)


zachdunn:
The double session resolution flagged in the review is fixed in e8503fab: `forwardToWorkspaceFiles` now records the already-resolved userId in an in-process `WeakMap<Request, string>` that `dualWorkspaceAuth` consults before falling back to `sessionAuth`. A WeakMap rather than a header because a header on an inbound request is indistinguishable from one an external caller forged — the map is only populatable in-process, so requests arriving via the public `/v1/workspaces/...` mount can never carry a pre-resolved identity. Membership checking still runs on every request. New tests assert exactly one `get-session` call on a forwarded alias, the WeakMap-miss path still authenticates, and membership rejection is unchanged (1559 tests green).

zachdunn:
Rebased onto main (now containing #614) and folded `files/by-path` into the canonical surface as promised: the handler moved to `workspace-files.ts` as `GET /:workspace/files/by-path` (dual-auth, next to facets/search) and `me.ts` keeps a one-line alias — exactly the two-line pattern this phase established. #614's route tests pass unchanged against the alias path, plus new bearer- and session-credential coverage on the canonical path. 1569 tests green, typecheck clean.

Issue #419, opened 2026-07-22
Render before/after pairs side by side in the managed comment


zachdunn:
## Problem

When a PR's attachments include a before/after pair, the managed comment renders them as independent images in upload order. The reader does the pairing in their head. Seen in practice on PR #417's own attachments comment: `hero-before` and `hero-after` render stacked, not side by side.

The data to pair them already exists: #365/#370 put `path` and `state` metadata into the comment pipeline (`--state before|after|empty|error|loading` is the documented capture habit), and the skills teach agents to tag both sides.

## Proposal

In `gatherCommentBody`'s rendering: when two attachments form a pair, render them as one side-by-side row (two-column table cell or equivalent GitHub-markdown-safe layout), labeled before / after.

Pairing rule, in order:
1. Same `path` metadata, one `state=before` and one `state=after`.
2. No path: filename stems that differ only by a `before`/`after` token (`hero-before.webp` / `hero-after.webp`, `before.png` / `after.png`).

Unpaired attachments render exactly as today. A pair where one side updates keeps the pairing (keys are stable per filename, so re-captures replace in place).

## Notes

- GitHub markdown constraint: side-by-side generally means a `<table>` or `| |` markdown table with images — verify rendering on github.com including mobile width before committing to a layout.
- Same rendering applies everywhere the comment is produced (client sync, webhook promote, reconcile) since they share the render path — no per-trigger work.
- Sequenced states beyond the pair (`empty`/`error`/`loading`) stay unpaired in v1.

Related: #365 (deferred this grouping), #370, the file-page counterpart view issue (filed alongside this one).

zachdunn:
File-page counterpart view: #420

PR #620, opened 2026-08-10
fix(cli): fold --state into the derived screenshot name and warn on overwrite

### LINKED ISSUE #618: uploads screenshot keys off the URL, so a same-URL before/after pair silently overwrites itself (and can never match #419's pairing rule 1)
## Problem

`uploads screenshot <url>` derives the object key from the captured URL. `--state` is
carried as metadata only, never as part of identity. So the canonical before/after
capture flow — the same route, twice, before and after a change — **silently
overwrites itself**:

```bash
uploads screenshot http://localhost:3100/docs/mcp --full-page --out before-mcp.png --state before
# ...make the change...
uploads screenshot http://localhost:3100/docs/mcp --full-page --out after-mcp.png --state after
```

Both calls return the identical key, and the second overwrites the first:

```
gh/buildinternet/sunny/branch/claude-docs-snippets-language-af0856/localhost-docs-mcp.webp
```

Observed on `uploads` (current npm) while preparing buildinternet/sunny#1009. Exit
code `0`, no warning, no `hint`. The failure is invisible at capture time — I only
caught it because I recognized the returned URL from the previous call. An agent
that captures "before", makes its change, captures "after", and pastes both URLs
into a PR body produces a before/after table where **both cells show the after
image**, and nothing anywhere reports a problem.

## Why this is worse than a generic key collision

It defeats the primary pairing rule of the before/after feature itself. From #419:

> 1. Same `path` metadata, one `state=before` and one `state=after`.

Two captures of the same URL are exactly that case — same auto-derived `path`, one
`before`, one `after`. But they can never coexist as two objects, because the key
that `path` is derived from is the same key. **Rule 1 is unreachable through the
one command that derives `path` automatically.** Only rule 2 (filename stems) is
reachable, and only via `put`, which keys off the filename.

## Why documentation doesn't save you

The github-screenshots skill actively steers toward this:

> Pass `--state before`/`--state after` and `--meta path=/route` as a habit — both,
> every time. Before/after is the whole point of most PR screenshots…

…which reasonably reads as "state participates in identity". Its worked example
happens to dodge the collision by mixing commands:

```bash
uploads put ./step1-before.png --meta path=/settings --state before
uploads screenshot http://localhost:4321/settings --out step2-after.png --state after
```

That's `put` (filename-keyed) for one side and `screenshot` (URL-keyed) for the
other. Nothing says the mix is load-bearing, so the natural simplification — use
`screenshot` for both sides — is the broken path. `screenshot --help` documents the
key layout accurately (`…/branch/<branch>/<name>`) but never says `<name>` comes
from the URL, nor that `--state` doesn't disambiguate.

## Suggested fixes

Roughly in order of preference; (1) alone would have saved me:

1. **Warn on overwrite.** When `screenshot`/`put` would replace an existing object
   whose `state` metadata differs from this call's `--state`, emit a stderr warning
   and a JSON `hint` — the same shape as the existing "add `--meta path=`" nudge.
   Cheap, and catches the general case (any silent clobber of a differently-tagged
   object), not just before/after.
2. **Fold `state` into the derived name** when `--state` is passed and the name is
   auto-derived — `localhost-docs-mcp-before.webp` / `-after.webp`. Distinct objects,
   rule 1 and rule 2 both match, re-captures of the same side still replace in place.
   Backwards-incompatible for anyone hard-coding auto-derived URLs, so possibly
   opt-in or next-major.
3. **Document it** in `screenshot --help` and the github-screenshots skill: say the
   key comes from the URL, that `--state` is metadata only, and that a same-URL
   before/after pair needs distinct `--out` files uploaded with `put`.

Happy to send a PR for (1) or (3) if you'd like — say which.

## Workaround (for the skill, if nothing else changes)

Capture to distinct local files, then upload by filename, from inside the repo so
staging still keys off the branch:

```bash
uploads screenshot <url> --out before-x.png --state before --no-upload
# ...change...
uploads screenshot <url> --out after-x.png  --state after  --no-upload
uploads put ./before-x.png ./after-x.png --meta path=/route
```

If a collision already happened, `uploads delete <key>` the clobbered object before
opening the PR — otherwise it auto-promotes into the managed comment mislabeled.

zachdunn:
Closes #618.

`uploads screenshot` derived the object name from the captured URL only, so the canonical before/after flow — same route, twice, with `--state before` then `--state after` — collided on one key and the second capture silently overwrote the first. `--state` was metadata only, which also made #419's pairing rule 1 unreachable through the one command that derives `path` automatically.

## Changes

**State folds into the derived name** (fix 2 from the issue). `captureScreenshot` runs the URL-derived filename through a new `foldStateIntoFilename` helper: `localhost-docs-mcp.png` + `--state before` → `localhost-docs-mcp-before.png`. Folding happens once at capture time, so every key layout (branch-staged default, `--pr`/`--issue`, `--ref`/`--prefix` dated) inherits it. It is idempotent, skipped entirely for an explicit `--key`, and re-capturing the same URL + state still replaces in place. The MCP `screenshot` tool already accepted `state` for metadata; it now folds identically for CLI/MCP parity. `put` is untouched — its name comes from a real user file.

No compat gate: for auto-derived names + `--state`, the old behavior was silent data loss, and same-side re-captures keep their replace-in-place idempotency.

**`screenshot` surfaces replacement** (fix 1). Human mode now prints the same `>> replaced existing object (same URL)` note `put` prints (`writeReplacedNote` is now shared), and `--format json` fills its single `hint` slot when a `--state` capture replaced an existing object — lowest priority behind the binding warning and staging note.

**Docs** (fix 3). `screenshot --help` states the name comes from the URL and how `--state` folds in; the github-screenshots skill's worked examples use `screenshot` for both sides of a same-URL pair (the `put` path stays as the existing-file alternative); the uploads-cli skill gains a "Key derivation and `--state`" section covering folding, idempotent re-capture, the `--key` exemption, and the replaced note / JSON fields.

## Testing

- New coverage: `foldStateIntoFilename` unit tests; state folded into derived keys (before/after produce distinct keys); no folding with `--key`; no double-append; replaced note in human/quiet modes; JSON `hint` on replaced + state — for both the CLI command and the MCP tool.
- Full root suite: 274 files, 3873 tests, all passing. `pnpm --filter @buildinternet/uploads typecheck` clean.
- Changeset: `@buildinternet/uploads` minor (verified against the changeset ignore list).

### DIFF
--- a/.changeset/screenshot-state-derived-name.md
+++ b/.changeset/screenshot-state-derived-name.md
@@ -1,0 +1,5 @@
+---
+"@buildinternet/uploads": minor
+---
+
+`screenshot` now folds `--state` into the derived object name (e.g. `app.example-settings-before.png` / `-after.png`), so capturing the same URL with different states no longer silently overwrites the earlier capture (issue #618). Folding is skipped when an explicit `--key` is given. `screenshot` also now prints the same `>> replaced existing object (same URL)` note `put` prints when an upload replaces an existing object, in both human output and as a `hint` in `--format json`.
--- a/packages/uploads/src/commands.ts
+++ b/packages/uploads/src/commands.ts
@@ -426,7 +426,7 @@ function formatOptimizeNote(opt: {
   return undefined;
 }
 
-function writeReplacedNote(
+export function writeReplacedNote(
   replaced: boolean | undefined,
   quiet: boolean,
   dryRun = false,
--- a/packages/uploads/src/commands/screenshot.ts
+++ b/packages/uploads/src/commands/screenshot.ts
@@ -27,6 +27,7 @@ import {
   putStagingNoteText,
   resolveStageBindingWarning,
   mergeStagingMeta,
+  writeReplacedNote,
   type BranchTarget,
 } from "../commands.js";
 import { resolvePutDefaults } from "../config.js";
@@ -87,6 +88,15 @@ they reference via file:// or relative paths only resolves with --via local).
 localhost/private-network URLs are reachable only by the
 local backend — with --via remote (or auto falling back to remote) these
 fail fast with a clear error instead of sending a doomed request.
+
+The object name is derived from the target URL (host + path), not chosen by
+you — e.g. https://app.example/settings becomes app.example-settings.png.
+--state folds into that derived name (a -before/-after/... suffix), so
+capturing the same URL with --state before then --state after produces two
+distinct objects instead of the second silently overwriting the first.
+Re-capturing the same URL + --state replaces that object in place — the
+intended idempotency for repeat captures. --key bypasses all of this and
+sets the whole object key verbatim (no folding).
 
 After capture, screenshots share the put upload pipeline: optional --frame,
 optimize-by-default, --pr/--issue attachment + --comment, --gallery, --meta.
@@ -456,6 +466,9 @@ export async function runScreenshot(
     reducedMotion,
     evalJs,
     initScript,
+    // Skip folding when an explicit --key was given — --key sets the whole
+    // key, so there's no auto-derived name to fold state into.
+    state: keyHint ? undefined : explicitMeta.state,
     measureSelectors: annotateSelectors.length > 0 ? annotateSelectors : undefined,
     apiUrl: ctx.config.apiUrl,
     token: ctx.config.token,
@@ -587,6 +600,7 @@ export async function runScreenshot(
         `>> optimized ${prepared.originalBytes} → ${prepared.outputBytes} bytes\n`,
       );
     }
+    writeReplacedNote(result.replaced, ctx.quiet, dryRun, result.wouldRefuse);
     process.stderr.write(`>> key: ${result.key}${dryRun ? " (dry run — not uploaded)" : ""}\n`);
     if (stagingTarget !== undefined) {
       process.stderr.write(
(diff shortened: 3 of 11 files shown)


zachdunn:
Pushed a hardening commit (ce6db6c) from a pre-merge review:

- **Fold only canonical state values.** The fold read the merged metadata bag's `state`, which a free-form `--meta state=…` (or the MCP `metadata` record) can populate with any printable ASCII including `/` and spaces — `validateMetaMap` doesn't restrict it to the enum, so such a value could have entered the derived object key. `foldStateIntoFilename` now folds only `before|after|empty|error|loading`; anything else stays metadata-only, exactly as before this PR. Covered by new unit + CLI-wiring tests (`--meta state=x/y` keeps the unfolded name while still landing in metadata).
- **Softened the replaced+state JSON hint.** With folding in place, `replaced` + `--state` almost always means an intended same-side re-capture, so the hint now reads as informational ("expected for repeat captures of the same URL + state") instead of implying a problem.

Full suite still green (3875 tests) and typecheck clean.

Known transitional note for the release: objects captured with `--state` by an older CLI sit at unfolded keys; the first post-upgrade re-capture lands at the new folded key instead of replacing them, so a mid-flight PR can briefly list both until the stale key is deleted.

PR #626, opened 2026-08-11
feat(web): workspace settings toggle for GitHub attachment ingestion

zachdunn:
Follow-up to #623/#625: the workspace-level default for GitHub attachment ingestion (`githubIngestAttachments`) was fully wired in the API but had no control in the settings UI — the only way to set it was a raw PATCH. Zach went looking for it on the settings page and it wasn't there.

Adds an **Ingest GitHub attachments** row to the GitHub comment card on `/account/workspaces/:name/settings`, following the card's existing tri-state convention: `auto (off)` / `on` / `off` (this knob's auto default is off, unlike the other rows). Saves through the same `comment-settings` PATCH cycle; `auto` clears the field back to unset. Repos still override with `ingestGithubAttachments` in `.uploads.yml`.

Browser-verified on the local stack: save persists (`ingestGithubAttachments: true` in the GET after PATCH), reload shows the selection, and resetting to auto clears the stored field.

Workspace settings: Ingest GitHub attachments tri-state row

### DIFF
--- a/apps/web/src/lib/api-client.ts
+++ b/apps/web/src/lib/api-client.ts
@@ -1166,6 +1166,8 @@ export interface CommentSettings {
   showMetadata: boolean | null;
   linkToFilePage: boolean | null;
   note: string | null;
+  /** `null` (auto) defaults to `false` — unlike the other tri-state fields, which default to `true`. */
+  ingestGithubAttachments: boolean | null;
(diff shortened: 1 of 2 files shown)


zachdunn:
Real live example, post-merge test:

image

image


image

Issue #414, opened 2026-07-22
product: private-repo attachment privacy (guessable gh/ keys on public CDN)


zachdunn:
## In plain terms

Hosted files for PR/issue attachments live on the public CDN. Managed GitHub keys (`gh/<owner>/<repo>/…`) are **stable and guessable**, so attachments for private GitHub repos are not secret to anyone who can invent the URL—even when the file page is “unlisted” / visibility-gated.

This is a known product/security tradeoff (stable overwrite URLs + Camo embeds vs secrecy). It needs a deliberate design, not a silent key-shape change.

## Context

- Roadmap: `docs/roadmap.md` — “Private-repo embed privacy”
- Skill copy: `skills/github-screenshots` notes public CDN
- Related but distinct: #166 (docs: unlisted vs byte privacy language on public CDN workspaces)

## Options to evaluate (spike)

1. Non-guessable object keys with a stable overwrite map (preserve `--pr` ergonomics)
2. Signed / time-limited `embedUrl` while keeping durable `url` optional
3. Repo-scoped edge auth for private installations
4. Workspace policy: refuse public base URL for private-repo attachments without opt-in

## Acceptance criteria (for the spike)

- [x] Written decision: chosen approach + rejected alternatives
- [x] Compatibility story for existing PR comments / Camo caches
- [x] Threat model: who can discover URLs today vs after
- [x] Follow-up implementation issue(s) if spike accepts a path

## Out of scope for v1 of this issue

- Implementing full edge auth without a design doc

addisonlynch:
@zachdunn I came across this when setting up a screenshot harness for a private repo (solo project) and have an idea that might work. 

One constraint that I think rules out more of these options than it looks like: GH fetches every external image through Camo, server-side and anonymously. No viewer cookies, nothing to authenticate with. So any origin that isn't github.com has to be readable by a stranger or it doesn't render at all. 

That means 1 and 4 are obscurity rather than access control, and 2 doesn't survive Camo caching the fetch. 3 has no viewer session at the edge to check. 

Looks like #139 already went as far as this shape allows per the [note](https://github.com/buildinternet/uploads/blob/cf12b7c18d94eb4d84e16407b86931785945a282/apps/api/src/routes/public-files.ts#L138-L143) in public-files.ts (metadata-only enforcement, bytes still reachable off the public domain, real privacy needs a workspace with no publicBaseUrl. Which then can't embed)

# Idea

[...]

PR #641, opened 2026-08-12
feat: randomized URL prefixes for private-repo attachments (#631)

### LINKED ISSUE #631: feat: randomized URL prefix for private-repo attachments (capability URLs)
## In plain terms

Decision from #414: private-repo attachments get unguessable URLs, matching GitHub Camo's own security model (durable unguessable URL, no auth gate — the only shape that both renders inline and survives Camo's anonymous server-side fetching).

Shape: a randomized base id **per branch**, with stable filenames beneath it —

```
storage.uploads.sh/gh/private/:randomized_id/filename.png
```

Stable tails preserve everything that keys on them (`--pr` overwrites, before/after pairing, comment dedupe, by-path grouping); the random segment makes discovery infeasible; rotating the id is the revocation story.

## Context

- #414 — threat model and decision (supersedes its four spike options)
- #139 / `public-files.ts` metadata-only-enforcement note — where the previous approach stopped
- Branch-staged screenshots — per-branch scoping means the id is mintable at stage time, before a PR exists

## Design points

1. **Id generation & storage.** Random (never derived from repo/branch — derivation is guessability), ~128 bits, stored in a mapping keyed by `(owner, repo, branch)`. Where: D1 alongside existing attachment state.
2. **Trigger.** Applies when the target repo is private (known via the App installation / repo-link status). Public repos keep current stable keys, unchanged.
3. **Rotation.** An explicit operation: mint a new id for the branch, re-sync the managed comment (existing re-sync machinery). Old prefix dies at the CDN.
4. **Compat.** Prospective only — new uploads under the new model; existing attachments stay where they are.
5. **Surfaces.** CLI, hosted MCP, GitHub App comment paths all mint/resolve through the same mapping; `/f/` pages and metadata are unaffected (keys change shape, contracts don't).

## Acceptance criteria

- [ ] Private-repo attachments served under randomized per-branch prefixes
- [ ] Anonymous fetch of an unrotated URL works; guessing without the id is infeasible; rotation kills old URLs
- [ ] Stable-tail ergonomics verified: overwrite, pairing, dedupe, grouping
- [ ] Threat model documented (who can discover URLs; forwarded-URL caveat stated plainly)

## Out of scope

- Public-repo randomization (no benefit; not a good default)
- Migration of existing attachments
- The github-branch provider (see #630)

zachdunn:
## In plain terms

Attachments for private repos currently live at guessable keys: anyone who can name the repo can construct `gh/<owner>/<repo>/pull/<n>/<file>` and fetch it. This PR gives private-repo attachments unguessable URLs instead — a random ~128-bit prefix per branch (`gh/private/<32-hex-id>/…`) with the same stable filenames beneath it. The model matches GitHub Camo's own: a durable unguessable URL with no auth gate, which is the only shape that both renders inline and survives Camo's anonymous re-fetches. Decided in #414; closes #631.

- Applies automatically when the uploads GitHub App can see the target repo is private. No flags. Public repos, and repos the App can't see, keep today's keys byte-for-byte.
- The id is minted server-side per (owner, repo, branch) and stored in D1 — never derived from repo or branch. Issue attachments and ingested assets use a repo-level id.
- Stable tails preserve the ergonomics that key on them: `--pr` overwrites, before/after pairing, comment dedupe, and by-path grouping all survive. The managed comment gathers both key shapes, so existing attachments keep rendering.
- Rotation is the revocation story: `uploads github rotate-prefix` mints a new id, moves every object (bytes, D1 metadata, gallery/ledger rows), deletes the old copies, and re-syncs the managed comment. Old URLs 404 at origin. The sweep is resumable across previously-retired ids.
- Fail-open by design: if privacy can't be determined (older server, no App, lookup failure), uploads proceed at plain keys — an upload is never blocked. Unauthorized callers get answers indistinguishable from a public repo's, and no id is minted for them.
- This is a capability URL, not access control: possession of a URL grants read until the id is rotated, and the object's public metadata still names the repo/PR/branch to a URL holder. [docs/private-attachments.md](docs/private-attachments.md) states the threat model plainly.
- Prospective only — existing attachments stay where they are (Camo has already cached those URLs, so retroactive secrecy isn't achievable).

## How to try it

On a private repo with the uploads App installed:

```bash
uploads put ./after.png --pr 123        # key: gh/private/<id>/pull/123/after.png
uploads github rotate-prefix            # revoke the current branch's URLs
```

## Technical notes

- New D1 table `github_private_prefixes` (partial unique index = one active id per repo+branch; retired rows are tombstones the rotation sweep drains). Migration auto-applies on merge.
- One resolve surface: `POST /v1/:workspace/github/private-prefix` for the CLI; hosted MCP, promotion, webhook, and ingestion call the same service in-process. Repo privacy comes from `GET /repos/{repo}` with the installation token (KV-cached, webhook write-through); PR targets resolve to the head branch's id server-side.
- Rotation clears inheritance-donated metadata rows before renaming (the content-hash inheritance path would otherwise collide), and its D1 rewrites are workspace-scoped.
- Two commits fix a pre-existing foot-gun encountered along the way: raw NUL bytes in source files (`file-metadata.ts` and a new test helper) that made git/grep treat them as binary.
- Deferred with the trail in the threat-model doc: rotation consumes upload quota per moved object; `rotated: true` doesn't yet flag an incomplete sweep (`MAX_LIST_PAGES`/concurrent writers — the doc says to re-run); narrow `allowedKeyPrefixes`/`maxKeyDepth` workspace configs can reject private-shape keys.

## Test plan

- [x] `pnpm test` — 293 files / 4236 tests green (root runner)
- [x] `pnpm typecheck`, `pnpm check` clean
- [x] Unit: id mint/retire race-safety on the partial unique index (real migration SQL via SqliteD1)
- [x] Resolve matrix: no App / no installation / public / unauthorized / privacy-lookup failure → plain, no row minted, no oracle
- [x] Promotion: dual-sweep (plain + private staged), same-filename dedupe (private wins), public path byte-identical regression
- [x] CLI/MCP: private keys on put/attach/screenshot/staged/list; 404 degradation asserts byte-identical pre-#631 keys
- [x] Rotation: move+rename+delete with inheritable metadata present, resumable retired-id sweep, copy-failure leaves the old object, cross-workspace rows untouched
- [ ] Prod smoke after deploy: one private-repo `uploads put --pr`, confirm `gh/private/` URL renders via Camo, then rotate and confirm the old URL 404s


zachdunn:
Resolved the merge conflict with main and brought the branch up to date with the #613 canonical-surface refactors, plus a cleanup pass:

- **1d912b0b** — merge of main; the only conflict was a type-block collision in `client.ts`.
- **da40be52** — the private-prefix resolve/rotate routes now also mount on the canonical `/v1/workspaces/:workspace/github/*` vertical (token-only, `files:write`, mirroring canonical comment/promote), and the CLI calls the canonical paths like its sibling github requests post-#632. Old bearer paths unchanged, matching the other five github routes.
- **01a0a331** — simplification pass: mode-owning key-builder helpers replace the repeated plain-vs-private ternaries, the CLI's multi-prefix list plumbing collapses into one seam (first-prefix-cursor rule encoded once), promote/ingest share one fail-open resolve wrapper, comment gathering pages prefixes concurrently, and prefix minting is read-first (one D1 query steady-state).

Root suite 293 files / 4296 tests green; typecheck and lint clean.

Issue #308, opened 2026-07-20
web: surface the /f/ file-page link in the /me file browser


zachdunn:
Follow-up to #303 (PR #305). That PR made the authenticated `/me/workspaces/:name/files` listing return a server-computed `pageUrl` (the public `/f/<workspace>/<key>` file page) for public-URL objects. But the web UI doesn't consume it yet — `WorkspaceFileTable` (`apps/web/src/components/WorkspaceFileTable.tsx`) and its DTOs (`WorkspaceFile`/`WorkspaceFolderFile` in `apps/web/src/lib/api-client.ts`) don't parse or render a `pageUrl` field; the "copy link" action resolves the raw/signed storage URL, not a file-page link.

## Task
- Extend the web DTOs to carry `pageUrl` from the API response.
- Add an affordance in the file browser to open / copy the `/f/` file page (metadata, dates, video player) when `pageUrl` is present (public-CDN workspaces). Objects without `pageUrl` (BYO/private-only) keep current behavior.
- Decide placement (a row action / link on the filename vs the existing copy-link menu).

## Notes
- API side already done in #303 — this is purely `apps/web`.
- `pageUrl` is absent for workspaces without a public base URL; the UI must handle its absence gracefully (no broken link).

Issue #135, opened 2026-07-13
Standalone public file page: chrome-wrapped single-object view (/f/<ws>/<key>)


zachdunn:
## In plain terms

Give any stored object a stable, public, chrome-wrapped page at `https://uploads.sh/f/<workspace>/<key…>` — a wrapper around the raw R2 bytes that presents filename, type, size, and upload time, with the original file one click away. It is the **single-item sibling of the public gallery** (`/g/<id>`): same public-read posture, same URL resolver, same rendering discipline, keyed on one object instead of a curated collection.

Immediate motivation: clicking an image in the account file browser today dumps the user at raw file bytes. There is no page that presents one file with metadata and context outside a gallery.

## Architecture (corrected)

`apps/web` has **no R2/D1 bindings** — it cannot HEAD storage directly. Like the public gallery page, the file page fetches from a **new unauthenticated API endpoint**, `GET /public/files/:workspace/:key`. That endpoint — not the Astro page — is the real security surface (404-parity, no-enumeration, rate-limit, cache).

Endpoint behavior (mirrors `public-galleries.ts`):
1. Resolve the workspace record from the route.
2. Resolve the original URL via the public URL (Phase 1) → this doubles as the visibility gate: no `publicBaseUrl` ⇒ 404. Swap in the #123 signed-URL resolver when it lands.
3. HEAD the object for `size`, `contentType`, `uploaded`; 404 on missing.
4. Omit raw provenance metadata (unlike the authenticated HEAD).

## Security delta

Raw bytes are already fully public and unsigned (R2 public custom domain, no auth/expiry; the worker never proxies bytes). A wrapper page **adds no new access**. Net-new risks are disclosure/discoverability only, mitigated:

- **Existence oracle** — single-key lookup only, same 404 surface R2 already exposes.
- **Metadata amplification** — allowlisted fields only (name/type/size/uploaded); raw provenance suppressed.
- **Enumeration** — hard no. One object per URL. No listing variant exists.
- **Traversal / workspace-crossing** — `badKey` (`KEY_RE` + `.`/`..` rejection) gates the endpoint directly; resolution is always scoped to the named workspace's prefixed storage.

## Scope (Phase 1 — public by default)

- Astro `src/pages/f/[workspace]/[...key].astro`, `prerender = false`, catch-all key.
- New endpoint mounted at `/public/files`; fetched server-side by `apps/web`.
[...]

Issue #123, opened 2026-07-13
fix(web): open files from private and BYO workspaces


zachdunn:
## Problem

The account file browser resolves a selected object through `GET /me/workspaces/:name/file-url`. Today that route only uses the workspace's `publicBaseUrl`, so it returns `{ "url": null }` for a valid private or BYO bucket without a public custom domain—even when the workspace has bucket-scoped credentials that Files SDK can use to create a signed URL.

That makes browsing work but opening a selected file fail silently for a supported storage configuration.

## Desired behavior

Resolve a usable file URL according to the workspace's storage capabilities:

- Prefer the stable public URL when `publicBaseUrl` is configured.
- Otherwise use Files SDK to generate a short-lived signed download URL when the provider has the required credentials.
- Return a typed, useful error when neither path is available; do not return a successful response containing `null`.

Keep membership authorization, communal-workspace behavior, key validation, existence checks, and workspace prefix isolation unchanged.

## Acceptance criteria

- [ ] Shared-prefix R2 workspaces with `publicBaseUrl` still return the encoded public object URL.
- [ ] Private/BYO workspaces with signing credentials return a usable, short-lived signed URL.
- [ ] Binding-only workspaces without a public URL or signing credentials receive a typed error that the account UI can display.
- [ ] Tests cover all three capability combinations plus unauthorized membership, invalid keys, and missing objects.
- [ ] No provider-specific adapter or R2 binding access is introduced in route code; storage access remains behind `@uploads/storage` / `createStorage()`.

## Context

Follow-up to #122.

PR #661, opened 2026-08-14
fix(web): open file tiles and rows as real links

zachdunn:
Opening a file from the screenshots page or the workspace files tab went through `window.open` on a click handler, so it cost a blank tab and a redirect even when the destination was already known. The tiles and rows weren't links at all — no middle-click, no cmd-click, no hover preview, and a popup blocker could eat them.

## The three destinations

Only one of them is genuinely unknown at render time:

| Case | Destination | Known at render? |
|---|---|---|
| API supplied `pageUrl` | the `/f/…` file page (#308) | yes |
| Workspace has a public domain | same-origin `/f/` path (#135) | yes |
| Private / BYO workspace | short-lived signed URL (#123) | **no** — minted on demand |

So the first two now render a real `<a href target="_blank" rel="noopener noreferrer">`, and only the signed-URL case keeps the open-blank-tab-then-redirect hop (the blank tab is opened synchronously so the popup blocker attributes it to the click).

## Why not a redirect endpoint

The obvious way to make *every* tile an anchor is a server-side route that 302s to the signed URL. I decided against it: it adds an auth-gated GET that redirects to a capability URL — a route that has to be reasoned about in every future security pass — and it's cross-origin (`api.` → `uploads.sh`), so it leans on SameSite behaviour on top-level navigation. That's real surface area to buy anchors for the minority path, when public workspaces are the common case. Worth revisiting if private workspaces ever become the norm.

## What changed

- **New `apps/web/src/lib/file-opener.ts`** — the resolution logic was about to exist in three copies, so it's one module now: `fileHref`, `makeFileOpener` (`href` + `activate`), `newTabLinkProps`.
- **`ScreenshotsByPath.tsx`** — `ShotThumb` takes an `href` and switches element; the `onOpen` prop threaded through the section components became a single `opener`.
- **`WorkspaceFileTable.tsx`** — a local `FileOpenTrigger` does the switch at all three open points (list-view name cell, grid card media, grid card name). Its `resolveSignedFileUrl` stays, because copy-link needs the raw URL for the clipboard rather than a navigation.

## Reviewer notes

Two element-qualified CSS selectors needed care, and one of them would have shipped as a visual bug:

- `button.wft-name--btn` would have **stopped matching entirely** once the element became an anchor, dropping the whole button reset. De-qualified.
- `.wft-card__name`'s hover rule is now scoped to `a, button` rather than de-qualified — a *folder* card's name is a plain `<span>` inside the card button, and dropping the qualifier would have given folder names an accent hover they never had.

**`AccountFileBrowser` is deliberately untouched.** It doesn't own its row markup — it renders `FileBrowser` from `@uploads/ui`, whose only opening API is `onSelect?: (file) => void`. Anchors there mean adding an href-per-item concept to a DS component with other consumers; that deserves its own change.

No changeset: `apps/web` only, and changesets cover the `@buildinternet/uploads` package (AGENTS.md).

## Verification

Local stack with a signed-in session, seeded `path`-tagged files, all four states checked by reading the rendered markup:

| Surface | Public workspace | Private/BYO |
|---|---|---|
| Screenshots overview + drill-in tiles | `A` → `/f/dev-demo/screenshots/….png` | `BUTTON type="button"` |
| Files list-view name cell | `A` | `BUTTON` |
| Files grid card (media + name) | `A`, both triggers | `BUTTON`, both triggers |

Folder cards stay `SPAN` throughout, and I confirmed the hover rule matches file-name anchors and not folder spans. Computed `text-decoration: none` and inherited colour on every converted element, so there's no visual diff. 43 web test files / 611 tests pass.

### DIFF
--- a/apps/web/src/components/ScreenshotsByPath.tsx
+++ b/apps/web/src/components/ScreenshotsByPath.tsx
@@ -19,8 +19,7 @@ import {
 import { loadWorkspaces } from "../lib/workspaces-nav";
 import { resolveWorkspaceInfo, type WorkspaceInfoStatus } from "../lib/workspace-file-row";
 import { onSession } from "../lib/account-shell";
-import { filePath } from "../lib/public-file";
-import { fetchWithTimeout } from "../lib/request";
+import { makeFileOpener, newTabLinkProps, type FileOpener } from "../lib/file-opener";
 import {
   lastUpdatedLabel,
   pairedShotKeys,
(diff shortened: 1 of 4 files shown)


zachdunn:
Changeset bot notice is a no-op here — no changeset is correct for this PR, and the link it offers would break the next release.

`@uploads/web` is `private: true` and sits in the `ignore` list in `.changeset/config.json`. Per AGENTS.md ("Releasing `@buildinternet/uploads`"): a changeset naming a private `@uploads/*` package yields an empty version PR and blocks the next publish. The bot's "add a changeset" link pre-fills exactly that (`"@uploads/web": patch`), so it should not be used.

Changesets here are only for user-visible `@buildinternet/uploads` (CLI/client/MCP) changes. This PR touches `apps/web` only.

PR #695, opened 2026-08-18
perf: Server-render signed-in page data to remove layout jump

zachdunn:
## What & why

Signed-in pages under `/account/*` are already on-demand rendered (`prerender = false`), but until now only `account/developers.astro` fetched its data *during* the server render. The rest painted a placeholder skeleton and fetched their primary data **client-side** after the session gate — the source of the skeleton→content jump and the slower-feeling load on those pages.

This PR propagates the proven `developers.astro` pattern (fetch with the request cookie in the frontmatter, render real content, let the client script revalidate) to the highest-value pages, and moves the post-login workspace resolution server-side.

## Changes

- **Cookie forwarding** (`api-client.ts`): optional `{ cookie }` on the `/me/*` helpers (`getMyWorkspaces`, `getWorkspaceSummary`, `getWorkspacePeople`, `getMyWorkspaceFiles`, `getWorkspaceFilesByPath`, and the shared `fetchWorkspaceList` behind `getMyWorkspaceGalleries`), routed through the existing `sessionFetchInit`. Enables server-side authenticated fetches. No change for existing browser callers — the option is optional.
- **Galleries tab**: server-renders the real table when the viewer is a member with ≥1 gallery; placeholder fallback otherwise. Also collapses the client summary→galleries waterfall on the server.
- **People tab**: server-renders the member rows read-only (`canManage: false`); the client re-render adds identity, per-row controls, and invites. Rows are identical across the two renders, so there is no layout jump.
- **Post-login landing**: `/account/workspaces` now 302s a single-workspace user straight to their workspace server-side (honoring `?to=screenshots`), removing a client fetch + JS boot + redirect hop before reaching content. Multi-workspace, `?manage=1`, empty, and unavailable states all fall through to the existing client script unchanged.

## Verification

- `pnpm --filter @uploads/web typecheck` — 0 errors
- `pnpm --filter @uploads/web test` — 644/644, including 12 new cookie-forwarding tests. They are load-bearing: reverting the source fails exactly the 6 "forwards" cases (assertions run outside the `fetch` mock so `fetchWithTimeout`'s catch can't swallow them).
- `pnpm lint` — clean
- Runtime (local raw stack, signed-in demo session):
  - **Galleries / People**: real content is present in the **server** HTML only with a valid session cookie; without one, the safe placeholder renders (no data leak).
  - **Landing**: `302 → /account/workspaces/<ws>` for a single-workspace user; `?to=screenshots` → `.../screenshots`; `?manage=1` and signed-out both render `200` with no redirect loop.

## Notes

- **No changeset** — `@uploads/web` is private; a changeset naming it would yield an empty version PR and block the next CLI publish.
- **Warm vs cold**: the SSR content paints immediately on warm SPA navigations (the common case). Cold first loads still wait behind the client session gate; removing that is a deeper change, tracked as a follow-up.
- Part of a broader SSR-first audit. The workspace **files** and **screenshots** tabs (React islands) are the remaining, higher-effort candidates and are intentionally out of this PR.

## Summary by CodeRabbit

* **New Features**
  * Workspace pages now load summaries, galleries, and member information on the server when possible, reducing empty states and improving initial page display.
  * Signed-in users with one workspace are redirected directly to it, while preserving the selected screenshots tab.
  * Workspace member listings initially include read-only role information before interactive controls load.

* **Bug Fixes**
  * Improved handling of authenticated workspace requests so session information is consistently retained across workspace views.

### DIFF
--- a/apps/web/src/lib/api-client.test.ts
+++ b/apps/web/src/lib/api-client.test.ts
@@ -3,6 +3,8 @@ import {
   deleteWorkspaceFile,
   deleteWorkspaceStorage,
   getGithubInstalled,
+  getMyWorkspaceFiles,
+  getMyWorkspaceGalleries,
   getMyWorkspaces,
   getSuggestedWorkspaceName,
   parseIssuedWorkspaceTokens,
@@ -12,7 +14,9 @@ import {
   getWorkspaceFilesByPath,
   getWorkspaceInvites,
   getWorkspaceMembers,
+  getWorkspacePeople,
   getWorkspaceStorageStatus,
+  getWorkspaceSummary,
   inviteToWorkspace,
   listWorkspaceFolder,
   putWorkspaceStorage,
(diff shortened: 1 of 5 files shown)


zachdunn:
@coderabbitai review

Issue #754, opened 2026-08-22
Pre-tagging simplification: merge auth D1, drop Secrets Store, minimal-binding profile


zachdunn:
## Why

#753 (platform releases for self-hosters) freezes the operator-facing surface: every worker, database, binding, and secret becomes part of a versioned compatibility contract with upgrade-notes obligations. Today the self-hoster count is ~zero, so structural simplification is as cheap as it will ever be. This issue tracks the cleanup that should land **before** the first `platform-v*` tag; afterwards, each of these becomes a hard-stop release instead of an internal migration.

## 1. Merge the auth D1 database into the main database — **decided: do it**

Two D1s means two provisioning steps, two migration chains, two backup/recovery stories, and `{api, auth}` pairs threaded through the release manifest, notes template, doctor command, and CI validation forever. The dedicated auth worker does not require a dedicated *database* — it can bind the same D1 with Better Auth's tables alongside ours (Better Auth's default posture is tables-in-your-app-DB). Needless complexity; consolidate now.

Scope: unify the migration chains (auth's `apps/auth/migrations/` folds into the main history), one-time data move for our prod, collapse `d1-migrations-auth.yml` into `d1-migrations.yml`, root `migrate:d1` becomes the single canonical command. Sequence relative to #731 so we're not migrating a moving target.

## 2. Move auth's Secrets Store secrets to plain worker secrets

The store's drift-prevention rationale doesn't hold in practice:

- All four store-backed secrets (`UPL_BETTER_AUTH_SECRET`, `UPL_BETTER_AUTH_API_KEY`, `UPL_GITHUB_CLIENT_ID`, `UPL_GITHUB_CLIENT_SECRET`) have exactly one consumer — the auth worker. No propagation benefit.
- The genuinely shared secrets (`BILLING_INTERNAL_KEY` api↔auth, `GITHUB_APP_PRIVATE_KEY` api↔mcp) are already duplicated plain secrets outside the store.
- Auth is mixed anyway (its Stripe secrets are plain), and the `store_id` points at the account-level store shared across buildinternet repos — inherently non-portable for a self-hoster.

Move the four to plain secrets; delete the `secrets_store_secrets` blocks (both prod and preview). The two shared plain secrets remain a small, documented drift surface the future doctor command can check.

## 3. Define a minimal-binding profile (core vs optional)

[...]

PR #756, opened 2026-08-22
feat(auth): dual-read plain secrets ahead of Secrets Store removal (#754 item 2)

zachdunn:
Part of #754 item 2: move the auth worker off the account-level Secrets Store onto plain per-worker secrets. This PR is the **safe-to-merge transition** — code prefers a plain env secret and falls back to the existing `UPL_*` store binding when the plain value is unset, so nothing breaks before the operator sets the plain secrets. Store bindings in `wrangler.jsonc` are deliberately kept until the follow-up removal PR.

## Changes

- `apps/auth/src/secrets.ts` — `resolveSigningSecret`, `resolveGitHubCredentials`, `resolveDashApiKey` now prefer `BETTER_AUTH_SECRET`, `GITHUB_CLIENT_ID`/`GITHUB_CLIENT_SECRET`, `BETTER_AUTH_API_KEY` (plain), falling back to the `UPL_*` store bindings.
- `BETTER_AUTH_SECRET_DEV` renamed to `BETTER_AUTH_SECRET` across env types, `.dev.vars.example`, docs, and 15 test fixtures — the other three plain names already existed as dev-only fallbacks and simply became the primary path.
- `secrets.test.ts` rewritten to assert the priority order (plain wins, store is fallback, per-field independence) plus the original store-failure/empty coverage; the 503-without-signing-secret behavior is unchanged and still covered.
- `docs/ops.md` gains the cutover runbook; `wrangler.jsonc` secrets comment rewritten (blocks untouched — also keeps this PR conflict-free with #755's D1 work).

## Operator cutover (after merge, no redeploy needed)

```
# from apps/auth, prod
pnpm exec wrangler secret put BETTER_AUTH_SECRET
pnpm exec wrangler secret put BETTER_AUTH_API_KEY
pnpm exec wrangler secret put GITHUB_CLIENT_ID
pnpm exec wrangler secret put GITHUB_CLIENT_SECRET
```

Then verify `/api/auth/*` (magic link + GitHub sign-in). Follow-up PR afterwards: delete both `secrets_store_secrets` blocks and the `UPL_*` fallback paths/types.

## Verification

- Full `pnpm test`: 314 files / 4748 tests green; `@uploads/auth` typecheck exit 0
- `format:check`, `changeset:lint` pass (no changeset — private package)
- Pre-existing `pnpm run lint` failures in web/api/mcp exist on main too (missing generated worker types) and are untouched here

Part of #754.

### DIFF
--- a/apps/auth/.dev.vars.example
+++ b/apps/auth/.dev.vars.example
@@ -1,11 +1,10 @@
 # Copy to .dev.vars for local dev.
 
-# Dev-only signing secret fallback. The Secrets Store binding
-# (UPL_BETTER_AUTH_SECRET) exists locally too under `wrangler dev`, but an
-# unpopulated store entry does NOT get overridden by a same-named .dev.vars
-# value (see src/secrets.ts) — that's why this is a distinctly-named var.
-# Without it, /api/auth/* answers 503 locally. Any string >= 32 chars works.
-BETTER_AUTH_SECRET_DEV=
+# Better Auth signing secret. Plain secret, preferred over the transitional
+# UPL_BETTER_AUTH_SECRET Secrets Store binding (see src/secrets.ts and
+# uploads#754 item 2). Without it, /api/auth/* answers 503 locally. Any
+# string >= 32 chars works.
+BETTER_AUTH_SECRET=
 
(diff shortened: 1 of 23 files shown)


zachdunn:
@coderabbitai review

Issue #556, opened 2026-07-29
chore(auth): upgrade better-auth to 1.7 and adopt CIMD client registration


zachdunn:
MCP spec `2026-07-28` deprecates [RFC 7591 Dynamic Client Registration](https://datatracker.ietf.org/doc/html/rfc7591) in favour of **Client ID Metadata Documents (CIMD)**, where `client_id` is an HTTPS URL pointing at a metadata document. This is Phase 2 of the MCP adoption work — Phase 1 (the protocol/SDK migration) is done separately and does not depend on this.

Sibling project `sunny` tracks the same migration in [buildinternet/sunny#778](https://github.com/buildinternet/sunny/issues/778); this issue is the uploads-specific version, audited against our own code rather than copied.

## Why we care

DCR is on by default for us — `allowUnauthenticatedClientRegistration: true` in `apps/auth/src/auth.ts:530` — and we run a **nightly stale-client reaper** (`apps/auth/src/oauth-client-reaper.ts`) precisely because anonymous DCR registrations accumulate as `oauth_client` rows from probes, MCP Inspector, and agents that register but never finish auth. CIMD creates no persistent client rows, so it removes that entire class of garbage rather than sweeping it up.

Not urgent: DCR remains spec-valid for backward compatibility, with a minimum twelve-month deprecation window under the new feature-lifecycle policy.

## Gated on better-auth 1.7

CIMD shipped upstream in [better-auth/better-auth#9159](https://github.com/better-auth/better-auth/pull/9159), releasing in **v1.7.0**. We run `better-auth ^1.6.23` and `@better-auth/oauth-provider ^1.6.23`, so adopting CIMD means taking the 1.6 → 1.7 upgrade first.

## What 1.7 actually costs us (audited 2026-07-29)

Where our exposure differs from sunny's, it is noted:

- [ ] **Account identity rework** ([#10403](https://github.com/better-auth/better-auth/pull/10403)) — `Account.accountId` → `Account.providerAccountId`, and `Account.issuer` becomes required. Hits `apps/auth/src/schema.ts:105` (`accountId: text("account_id")`) and the client surface in `apps/web/src/lib/auth-client.ts`, which reads `accountId` from `listAccounts` in several places. Needs an additive D1 migration plus a reviewed backfill — the upstream guide is explicit that trusted issuers cannot be assigned automatically.
[...]

PR #760, opened 2026-08-22
feat(auth): adopt CIMD client registration alongside DCR (#556)

zachdunn:
Towards #556. The blocker (Better Auth 1.7) landed in #744, so this enables the CIMD track itself.

## What changed

- **`@better-auth/cimd@1.7.1` registered on the auth worker** with the `mcp-2026-07-28` metadata profile (the MCP revision pins CIMD draft-00: `client_name` + `redirect_uris` required). An MCP client can now use an HTTPS URL as its `client_id`; the worker fetches, validates, and persists it as a discovery-owned `oauth_client` row (`client_discovery_id = 'cimd'`). AS metadata now advertises `client_id_metadata_document_supported: true`.
- **Workers fetch transport** (`apps/auth/src/cimd-transport.ts`) — the package's own transport needs `node:dns`/`node:https`. Ours enforces HTTPS, GET/HEAD only, draft-02 URL validation (loopback/private/link-local/cloud-metadata hosts rejected up front), and `redirect: "manual"`. The resolve-once-and-pin DNS guarantee is delegated to the Cloudflare edge network boundary, where private address space is not routable to any uploads service — the reasoning is spelled out in the module comment.
- **DCR stays on** for backward compatibility (the spec keeps it valid with a ≥12-month window). The existing register test plus a new one confirm the handler tolerates SEP-837's `application_type`.
- **Stale-client reaper stays** (issue expected it might retire): Better Auth's CIMD implementation *does* persist client rows, so anonymous URL-form clients accumulate exactly like DCR rows. They are deliberately not exempted — a swept CIMD row is recreated on demand from its metadata URL, and the never-used guard protects rows with live consents/tokens.
- **Docs**: `auth.md` and `llms-full.txt` now describe CIMD-first registration with DCR as the compatibility path.

## Issue checklist coverage

- [x] Enable CIMD client resolution alongside DCR
- [x] `redirect_uris` matching — covered in tests: exact-match failure 302s to the AS's own error page (`error=invalid_redirect`), never the unlisted target; real-client verification against `agents.uploads.sh` still owed post-deploy
- [x] DCR tolerates `application_type` (SEP-837)
- [x] Docs updated for CIMD-first registration
- [x] Reaper re-evaluated (kept, with rationale in its header)

## Verification

- 4757/4757 repo tests (9 new in `apps/auth/src/cimd.test.ts`, driven through the real handler + fake-D1, metadata fetch stubbed at the global `fetch` seam)
- Repo typecheck 0 errors (Node 24), lint clean
- `wrangler deploy --dry-run` bundles cleanly (the infra-0.4 lazy-import class of breakage that tsc/vitest miss)

## Owed after merge

Prod smoke with a real CIMD-speaking MCP client against `agents.uploads.sh` (the known `redirect_uris` ecosystem sharp edge), then a follow-up comment on #556.

### DIFF
--- a/apps/auth/package.json
+++ b/apps/auth/package.json
@@ -13,6 +13,7 @@
     "test": "vitest run"
   },
   "dependencies": {
+    "@better-auth/cimd": "^1.7.1",
     "@better-auth/infra": "^0.4.1",
     "@better-auth/oauth-provider": "^1.7.1",
     "@better-auth/stripe": "^1.7.1",
--- a/apps/auth/src/auth.ts
+++ b/apps/auth/src/auth.ts
@@ -5,6 +5,7 @@
  * after a redeploy, or a different D1 binding under `wrangler dev -c`) never
  * serves a stale instance.
  */
+import { cimd } from "@better-auth/cimd";
 import { dash } from "@better-auth/infra";
 import { oauthProvider } from "@better-auth/oauth-provider";
 import { drizzleAdapter } from "better-auth/adapters/drizzle";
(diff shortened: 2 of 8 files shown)


zachdunn:
@coderabbitai review

PR #800, opened 2026-08-23
feat(web,ui): shadcn Sidebar shell for signed-in layouts

### LINKED ISSUE #798: Evaluate shadcn Sidebar block for the signed-in shell
## Context

The Tailwind + shadcn migration now covers tables, inputs, dialogs, empty states, kbd hints, and switches. The one shadcn block we deliberately deferred is the [Sidebar](https://ui.shadcn.com/blocks/sidebar) for the signed-in shell.

## Why deferred

- The current shell (`apps/web/src/layouts/AccountLayout.astro` + `WorkspaceLayout.astro`) is pure Astro + vanilla TS, redesigned recently in #601 (contextual rail, workspace switcher, loading skeletons). It works and is fast.
- shadcn's Sidebar is React-only and depends on `sheet`, `separator`, `skeleton`, and `tooltip` — adopting it means converting the whole signed-in shell into a React island (mounted via `hydrateRoot`, since `client:*` directives are banned with the Cloudflare adapter).
- That is an architectural change to every signed-in page for little immediate user-visible gain.

## When to revisit

Worth a real design discussion once the current component adoption settles. Signals that would tip the balance:

- The workspace switcher or nav needs richer interactivity (search, pinning, keyboard nav) that vanilla TS makes painful.
- We want collapsible/mobile sheet behavior in the shell.
- More of the signed-in surface becomes React anyway, shrinking the marginal cost of a React shell.

Scope of the eventual work: port `AccountLayout`'s nav + rail into a shadcn Sidebar-based island, keep SSR-first rendering, preserve the #601 information hierarchy, and add the missing kit deps (sheet/separator/skeleton).

zachdunn:
Closes #798. The signed-in shells' left nav is now the shadcn Sidebar block, running as a React island in both the account and admin layouts — collapse-to-icons with ⌘B, cookie-persisted state with no SSR flash, a mobile Sheet drawer, and a workspace switcher rebuilt on kit `DropdownMenu`. Design spec agreed in-session and kept at `.context/2026-08-23-shadcn-sidebar-shell-design.md`.

## What's here

**Kit (`packages/ui`)**
- `Sidebar`, `Sheet`, `Separator`, `Skeleton` + the `use-mobile` hook via the shadcn CLI; tsup/exports wiring; new `--color-sidebar-*` theme tokens mapped onto the brand palette; UiKit workbench sections for all four.

**ShellSidebar island (`apps/web`)**
- One parameterized component for both shells: `components/shell/ShellSidebar.tsx`, SSR-rendered and `hydrateRoot`-ed exactly like the files table (no `client:*`). Props shaping lives in a pure module (`lib/shell-sidebar-data.ts`) with 20 unit tests.
- `SidebarProvider` scoped to the island; both layouts read the `sidebar_state` cookie server-side so a collapsed sidebar SSRs collapsed — no flash, verified across navigations.
- Workspace switcher on kit `DropdownMenu` in the SidebarHeader (workspace list + create link); admin gets a plain header label.
- Mobile (<768px): the island renders the kit's Sheet drawer; a small header button dispatches `shell:sidebar-toggle`, which the island listens for. The bespoke `.side-*` / `.ws-switcher` CSS and the old inline nav-painting script are deleted.
- Right rail, `workspace-rail.ts`, and all page bodies untouched, per the spec.

**Dev-server fix uncovered along the way**
- With `hmr: false`, Vite can't deliver the full-reload it issues after a cold-cache dep re-optimization, so the first signed-in page load after a cache wipe stranded every island on a duplicate React ("Invalid hook call"). The lazily-discovered deps are Astro's own ClientRouter virtual modules — now pre-included in `optimizeDeps`, plus `resolve.dedupe` for React. Cold-start verified clean.
- Also fixed in passing: an `is:inline` script wrapped in a JSX `{`…`}` expression renders as a literal no-op template string — the header toggle binder now ships via `set:html`.

## Screenshots

| Expanded | Collapsed (⌘B, icon rail) |
| --- | --- |
| Account shell with the expanded shadcn sidebar and workspace switcher | Same shell collapsed to the icon rail |

| Admin shell | Mobile sheet drawer |
| --- | --- |
| Admin shell on the same sidebar island | Mobile navigation as a sheet drawer |

## Verification

- `tsc --noEmit` clean; web vitest 868/868; repo `format:check` clean.
- Browser-verified on the local stack: hydration with zero console errors, ⌘B collapse + cookie round-trip, SSR persistence across navigations, switcher dropdown, mobile sheet via the header button at 375px, admin parity, right rail intact, cold-cache first load.

Follow-up candidates (not in this PR): retire the now-orphaned `workspaces-nav.ts` render helpers (`initWorkspacesNav`, `renderSwitcherMenuHtml`, `renderSettingsSubnavHtml`, …) and their tests; type-to-search Combobox switcher.


zachdunn:
Design pass per Zach's feedback: the active nav row now uses an accent-tinted surface with accent text (OpenRouter-style) instead of the neutral wash, and the shell's global link underline no longer applies inside the sidebar (covers the mobile sheet too, which portals to body). The kit sidebar carries its own hover/active affordances, so the underline was redundant there. The Expanded screenshot in the description reflects the new treatment (embed URL is stable).

zachdunn:
Two more design-feedback commits:

- **b578a13** drops the identity footer at the bottom of the sidebar — the header avatar menu already carries identity + sign-out, so the footer duplicated it. The `identity` prop plumbing is removed from both layouts.
- **b8958da** capitalizes the nav labels (Screenshots, Files, GitHub comment, …) and gives the Workspace/Personal group headings the same small-caps treatment as the right rail's labels.

The screenshots in the description have been re-captured in place to match.

Issue #731, opened 2026-08-21
Same-origin auth: host-only session cookie with uploads.sh as the front door and issuer


zachdunn:
Spun out of #684. That issue is about preview URLs; while revisiting it we concluded the underlying constraint — the `.uploads.sh`-scoped session cookie — deserves its own migration, independent of previews.

## Problem

The session cookie is `Domain=.uploads.sh` (`crossSubDomainCookies` in `apps/auth/src/auth.ts`, derived in `deriveCookieDomain`) because three hosts consume it today: `uploads.sh` (web pages + `/device` approval), `auth.uploads.sh` (credentialed XHR from `src/lib/auth-client.ts`), and `api.uploads.sh` (credentialed XHR to `/me/*` from `src/lib/api-client.ts`). RFC 6265 offers no way to enumerate hosts — a cookie is either host-only or covers *every* subdomain. So today the production session rides along to any `uploads.sh` subdomain, including:

- **User-content origins** — `storage.uploads.sh` / `store.uploads.sh` / `embed.uploads.sh` serve raw user bytes (`apps/api/src/guards.ts` already notes the stored-XSS-on-own-origin risk).
- **Any future preview hostnames** under the zone (#684's blocker).

Prior art is one-sided: platforms either keep sessions host-only on the product origin (GitHub) or run a full IdP handoff across origins (Google, Stripe Connect, Okta). Nobody ships the parent-domain shared cookie. Stripe is the closest full parallel to our target: token-only `api.stripe.com`, dashboard session on its own host, browser talking only to its own origin (BFF), OAuth for anything on another domain, and scoped ephemeral sessions for satellite origins.

## Target end state ("GitHub shape")

- Session cookie is **host-only on `uploads.sh`** (eventually `__Host-`-prefixed). No other host ever receives it.
- `uploads.sh` is both the app origin and the OAuth issuer: **`https://uploads.sh/api/auth`**. Web's session and the AS session are the same Better Auth session — the product origin *is* the IdP, like GitHub.
- `api.uploads.sh` is **token-only** (workspace/admin/operator tokens, OAuth JWTs). No cookies, no credentialed CORS.
- The browser talks only to `uploads.sh`; auth and api are reached server-side over the existing service bindings (BFF pattern).
- `auth.uploads.sh` retires from the browser's view; it remains a deployment unit behind routing/bindings.

## Why now

[...]

PR #821, opened 2026-08-24
refactor(mcp): drop legacy pre-flip OAuth issuer acceptance (#731 phase E)

zachdunn:
Prod flip from **#731 phase E**. The MCP resource server accepted both the current `uploads.sh/api/auth` issuer and the pre-flip `auth.uploads.sh/api/auth` issuer, so OAuth access tokens minted before the phase-C cookie/issuer flip kept verifying. This drops the legacy acceptance.

## Why it's safe now
- OAuth access tokens use the plugin default (~1h) expiry; the phase-C flip landed **2026-08-21**, so no legacy-issued access token is still live.
- Refreshes already re-mint under the current issuer, and RFC 9728 discovery (`respondProtectedResource`) only ever advertised the current issuer.
- Worst case for a stale token: a `401` + `invalid_token` challenge → the client re-discovers the current issuer and re-authorizes once. Trivial for the current MCP user base.

## Change
- Removed `LEGACY_OAUTH_ISSUER` and the dual-issuer array in `oauthAuth`; `verifyOAuthJwt` now checks only the current issuer.
- Updated the now-stale dual-issuer comments in `oauth.ts` (the `issuer` array capability itself is retained as a general feature).
- Flipped the unit test that asserted legacy-issuer acceptance to assert a `401` + `invalid_token` challenge.

## Verification
`pnpm --filter @uploads/mcp test` — 120 pass. Typecheck clean.

Part of #731 phase E. Sibling flip (api credentialed CORS retirement) is a separate PR; `__Host-` cookie prefix remains parked.

### DIFF
--- a/apps/mcp/src/index.ts
+++ b/apps/mcp/src/index.ts
@@ -61,19 +61,6 @@ function authOriginOf(env: Env): string {
   return (env.AUTH_ORIGIN || "https://uploads.sh").replace(/\/+$/, "");
 }
 
-/**
- * The pre-#731-phase-C issuer — tokens minted before the cookie/issuer flip
- * still carry this `iss` and remain valid until they expire naturally.
- * `oauthAuth` accepts both this and the current issuer so those tokens keep
- * working through the migration window; RFC 9728 discovery
- * (`respondProtectedResource`) advertises ONLY the current issuer — new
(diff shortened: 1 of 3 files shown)


zachdunn:
@coderabbitai review

Issue #829, opened 2026-08-24
Improve the public API contract, retry safety, and scalability


zachdunn:
## Problem

The API implementation has strong compatibility, authentication, error, and pagination foundations, but the published contract and some operational semantics lag behind the current canonical surface.

The shipped client uses `/v1/workspaces/:workspace/...`, while the Markdown API reference and discoverable OpenAPI document still primarily describe the legacy `/v1/:workspace/...` routes. The OpenAPI document also disagrees with runtime behavior for some statuses, such as file deletion.

Retries and high-volume reads need a clearer contract too:

- a lost response after token minting can leave an unknown active credential
- a PUT can store bytes before later metadata work fails, so replaying it is not fully idempotent across quota and bookkeeping
- expensive authenticated reads have no dedicated rate limit
- most 429 responses do not include actionable retry metadata
- file search reports truncation without a continuation cursor
- canonical file listing always hydrates metadata, even when the client discards it

This follows the practical guidance in Sean Goedecke's [Everything I know about good API design](https://www.seangoedecke.com/good-api-design/): do not break userspace, make mutations safely retriable, bound API workloads, expose rate-limit recovery information, and use cursor pagination for growing collections.

### 1. Publish the canonical contract

- [x] Document `/v1/workspaces/:workspace/...` as the primary public surface.
- [x] Keep `/v1/:workspace/...` routes documented as compatibility aliases.
- [x] Expand OpenAPI to cover the canonical file, usage, gallery, and related public developer routes.
- [x] Correct response statuses and schemas, including file deletion.
- [x] Add contract tests that catch drift between registered public routes, the narrative reference, and OpenAPI.

### 2. Add retry-safe mutations

- [x] Define an optional `Idempotency-Key` contract scoped to the authenticated caller and workspace.
- [x] Persist a request fingerprint and replayable result for the retention window.
- [x] Start with token minting, uploads, and gallery creation.
- [x] Reject reuse of one key with a different request fingerprint.
- [x] Update the client to retry only when the operation carries an idempotency key.

### 3. Bound expensive reads and improve 429s

- [x] Add a separate per-workspace read limiter.
- [x] Use tighter limits for search, facets, and metadata-hydrated listings.
- [x] Emit standard `Retry-After` consistently.
- [x] Emit limit/remaining/reset metadata only when it is accurate.
- [x] Preserve separate write, render, poster, and public-intake quotas.

### 4. Make search pageable

- [x] Add an opaque search cursor.
[...]

PR #832, opened 2026-08-24
feat(api): make gallery creation retry-safe

zachdunn:
## In plain terms

Gallery creation can now survive a timeout or dropped response without creating a duplicate gallery. The API replays the original successful result when the caller retries with the same idempotency key.

Refs #829.

- Adds optional Idempotency-Key support to canonical and compatibility gallery-create routes.
- Scopes each key to the workspace, logical operation, and authenticated credential.
- Rejects a reused key when the effective request body changes.
- Generates an idempotency key in the JavaScript client and retries that POST once on network, 429, or 503 failures.
- Retains successful responses for 24 hours and removes expired records through the daily cleanup task.
- Documents the contract in the public OpenAPI document and API guide.
- Does not add token-mint or upload idempotency. Token replay needs encrypted secret storage. Upload replay needs recovery across D1 and object storage.

## How to try it

Send the same gallery body and Idempotency-Key twice:

```bash
curl -X POST "https://api.uploads.sh/v1/workspaces/$UPLOADS_WORKSPACE/galleries" \
  -H "Authorization: Bearer $UPLOADS_TOKEN" \
  -H "Content-Type: application/json" \
  -H "Idempotency-Key: gallery-demo-1" \
  -d '{"title":"Launch media"}'
```

The second response keeps status 201, returns the original gallery, and includes Idempotency-Replayed: true.

## Technical notes

- One transactional D1 batch claims the key, inserts the gallery, and stores the response.
- Idempotency reads use a primary-constrained D1 session so a retry cannot observe a stale replica.
- The database stores a SHA-256 digest of the raw key, not the key itself.
- The new migration is included in the branch. I did not apply migrations manually.

## Test plan

- [x] API test suite: 161 files, 2,240 tests
- [x] Client test suite: 74 files, 1,311 tests
- [x] Web test suite: 58 files, 883 tests
- [x] API, client, and errors TypeScript checks
- [x] Changed-file oxlint and oxfmt checks
- [ ] Full monorepo typecheck in this worktree; unrelated workspace dependencies were unavailable offline. CI will run the complete gate.

### DIFF
--- a/.changeset/safe-gallery-retries.md
+++ b/.changeset/safe-gallery-retries.md
@@ -1,0 +1,5 @@
+---
+"@buildinternet/uploads": patch
+---
+
+Retry gallery creation safely with caller-scoped idempotency keys.
--- a/apps/api/migrations/20260824120000_idempotency_requests.sql
+++ b/apps/api/migrations/20260824120000_idempotency_requests.sql
@@ -1,0 +1,17 @@
+CREATE TABLE idempotency_requests (
+  workspace TEXT NOT NULL,
+  principal TEXT NOT NULL,
+  operation TEXT NOT NULL,
+  key_hash TEXT NOT NULL,
+  fingerprint TEXT NOT NULL,
+  owner_nonce TEXT,
+  state TEXT NOT NULL CHECK (state IN ('pending', 'completed')),
+  response_status INTEGER,
+  response_body TEXT,
+  created_at TEXT NOT NULL,
+  expires_at TEXT NOT NULL,
+  PRIMARY KEY (workspace, principal, operation, key_hash)
+);
+
+CREATE INDEX idempotency_requests_expires_idx
+  ON idempotency_requests (expires_at);
--- a/apps/api/src/db-session.ts
+++ b/apps/api/src/db-session.ts
@@ -56,3 +56,12 @@ export function createDbSession(db: D1Database): D1DatabaseSession {
 export function dbFor(env: Env): D1Queryable {
   return env.DB_SESSION ?? env.DB;
 }
+
+/**
+ * Idempotency decisions must observe the latest claim from another request.
+ * Use a primary-constrained session even when ordinary reads may start on a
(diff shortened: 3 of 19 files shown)


zachdunn:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

zachdunn:
### Code review — follow-ups

Two lower-confidence notes (scored ~75) that fell below the auto-report threshold but are worth a look. Neither is blocking; both are consistency nits.

**1. Replay-lookup read is not bounded, unlike sibling gallery reads.** `createGalleryIdempotently` does its `db.batch(...)` write (correctly left unbounded), then falls through to a plain `.first()` SELECT to decide replay/conflict/limit. That is a pure read on the exact path a retry hits, but it is not wrapped in `boundedRead`/`boundedDataRead` the way every other read in the galleries vertical is (added in #816 so a D1 stall fails fast with a 503 instead of hanging). `gallery-idempotency.ts` does not take a Hono `Context` today, so threading `env`/`boundedRead` through would be the fix.

https://github.com/buildinternet/uploads/blob/6192c67e9a3b4aa07c59ba8a1d5a7f11d2c4d2bc/apps/api/src/gallery-idempotency.ts#L125-L134

**2. Cron cleanup passes `env.DB` directly instead of `dbFor(env)`.** The scheduled purge is the one D1 call site that bypasses the `dbFor(env)` access point the codebase documents as mandatory — index.ts itself says "Everything downstream reads D1 via `dbFor(env)` ... instead of `env.DB` directly". Behaviorally silent today, but the sibling cron jobs (`retention-sweep`, `observability-retention`) all go through `dbFor(env)`.

https://github.com/buildinternet/uploads/blob/6192c67e9a3b4aa07c59ba8a1d5a7f11d2c4d2bc/apps/api/src/index.ts#L259-L261

🤖 Generated with [Claude Code](https://claude.ai/code)

Issue #911, opened 2026-08-31
OAuth AS never issues refresh tokens: offline_access is not grantable


zachdunn:
Found while writing the regression test for #909 (PR #910).

`@better-auth/oauth-provider` only issues a refresh token when the grant's scopes include `offline_access` (`createUserTokens`: `isRefreshToken = ... && scopes.includes("offline_access")`). Our AS never grants it:

- `OAUTH_SCOPES` / DCR default+allowed scopes are `files:*` only (apps/auth/src/auth.ts), so no self-registered client is registered with it;
- the downscope hook (apps/auth/src/oauth-grant-scopes.ts, `KNOWN_SCOPES`) strips it from every authorize request;
- a refresh grant requesting it fails `client does not allow scope offline_access`.

Net effect: **every OAuth/MCP client (OpenCode, Cursor, Claude, MCP Inspector) gets a 1h access token and no refresh token.** When it expires the client's only option is a full interactive re-auth. This is very likely the primary cause of the "OpenCode needs re-auth within a day" report — #909's rotation reuse grace is correct but can't help until refresh tokens exist.

Fix sketch:
- add `offline_access` to `clientRegistrationDefaultScopes`/`clientRegistrationAllowedScopes` (and whatever plumbing lets it through `restrictOAuthQueryScopes`/consent — extend `KNOWN_SCOPES` or special-case user-delegated scopes);
- decide consent-screen copy for it ("stay signed in" style label on /oauth/consent);
- verify the connected-apps revoke flow (#890/#891) also kills refresh tokens;
- prod smoke test afterward: authorize via MCP Inspector, confirm `refresh_token` in the token response, refresh twice within 60s (replay), once after (rotation).

Related: #909, PR #910, and the observability follow-up issue for refresh-grant failures.

PR #915, opened 2026-08-31
feat(auth): observability for OAuth refresh-grant failures and rotation replays

### LINKED ISSUE #912: Observability for OAuth refresh-grant failures and rotation replays
Follow-up to #909 / #911. We had no signal that OAuth refresh grants were failing (or, per #911, never possible) — a user report was the detector.

Add lightweight analytics/log events in apps/auth for:

- refresh-token family invalidation (the RFC 9700 teardown path) — client_id + user, so a burst is visible;
- rotation replay hits (reuse within the 60s grace) — confirms #909 is actually absorbing concurrent refreshes once refresh tokens exist;
- refresh grant failures by error (`invalid_grant`, `invalid_scope`, ...) at `/oauth2/token`;
- token grants issued **without** a refresh token (would have made #911 obvious immediately).

The worker already writes to an Analytics Engine dataset in apps/api; apps/auth may need a binding. A daily glance (or /admin/metrics tile) beats waiting for the next user report. Server metrics won't catch client-side bugs like OpenCode holding a stale token in memory after re-auth (anomalyco/opencode#21702), but nonzero replay counts + zero family teardowns is the healthy signature to look for.

zachdunn:
## Summary

Closes #912. Follow-up to #909 (refresh-token rotation reuse grace) and #911 (offline_access grant). Before this, the only detector for a refresh grant silently failing, or a client never getting a refresh token at all, was a user report.

Adds an `AUTH_EVENTS` Analytics Engine binding to `apps/auth` (same pattern as apps/api's `ANALYTICS`/`SLOW_OPS` datasets) and a new `apps/auth/src/oauth-observability.ts` module wired into `betterAuth()`'s `hooks.before`/`hooks.after` at `/oauth2/token`. `@better-auth/oauth-provider` is a dependency and is never patched — this observes from outside via the hook pipeline plus reads of the `oauth_refresh_token` D1 rows the plugin already writes.

## Events emitted

All events write to the `uploads_auth_events` dataset (`uploads_auth_events_preview` in the previews block). Blob ordinal contract (`AUTH_EVENT_BLOB_ORDER` in oauth-observability.ts): `[event, clientId, grantType, errorCode, userId]`. Fields are minimal by design — client_id, an opaque userId, grant_type, and an OAuth error code where relevant. Never a token value or an email.

- **`refresh_family_invalidated`** — RFC 9700 §4.14 teardown: a revoked refresh token was reused *outside* the 60s reuse grace, so the plugin tore down the whole family and forced re-auth. Fields: `clientId`, `userId`, `grantType`.
- **`refresh_rotation_replay`** — a revoked refresh token was reused *within* the 60s reuse grace — confirms #909's absorption path is actually live now that refresh tokens exist (#913). Fields: `clientId`, `userId`, `grantType`.
- **`refresh_grant_failed`** — any other refresh-grant failure at `/oauth2/token`, by RFC 6749 error code (`invalid_grant` for an unknown/expired/wrong-client token, `invalid_scope`, `invalid_target`, ...). Fields: `clientId`, `grantType`, `errorCode`.
- **`token_grant_without_refresh`** — a successful `authorization_code` or `refresh_token` grant whose response carried no `refresh_token` — the exact shape #911 was, and would have made it obvious immediately. Fields: `clientId`, `grantType`.

The healthy signature to watch for (per the issue): nonzero `refresh_rotation_replay` counts with zero `refresh_family_invalidated`.

## How it works / cost

- `hooks.before` snapshots the presented refresh token's D1 row (revoked + rotation-replay-window state) **before** the plugin's own handler runs — the only extra query this adds, and only for `grant_type=refresh_token` requests at `/oauth2/token`. Every other request path (get-session, sign-in, other grant types, ...) is untouched.
- `hooks.after` reads the handler's outcome (`ctx.context.returned` — either the success JSON or the thrown `APIError`) plus that snapshot, classifies, and fires at most one AE point.
- The before→after snapshot handoff rides on `ctx.context` (Better Auth's shared per-request `AuthContext`), not `ctx` itself — `runBeforeHooks` in better-auth's dispatcher invokes the before-hook with a shallow-spread copy of `ctx`, so a field written directly on `ctx` would be discarded; `ctx.context` is a nested reference the spread preserves, and `runAfterHooks` passes the original context through directly. Verified against better-auth 1.7.1's `api/dispatch.mjs`; documented in the module so a future version bump that breaks this degrades to an unclassified event rather than misclassifying.
- `writeAuthEventPoint` never throws and never blocks — an absent `AUTH_EVENTS` binding (unit tests, local dev, self-hosters) is a silent no-op.

## How to query

Analytics Engine has no read binding — reads go through the SQL API (account-scoped, same as apps/api's dataset). Example:

```sql
SELECT blob1 AS event, blob2 AS client_id, blob4 AS error_code, count() AS n
FROM uploads_auth_events
WHERE timestamp > now() - INTERVAL '1' DAY
GROUP BY event, client_id, error_code
ORDER BY n DESC
```

## Testing notes

- New file `apps/auth/src/oauth-observability.test.ts` (20 tests): pure unit coverage of the classification logic (`classifyOAuthTokenEvent`) for every branch, `writeAuthEventPoint`'s no-op/write/never-throws contract, `snapshotPresentedRefreshToken` against the fake-D1 harness, and four end-to-end tests driving real `/api/auth/oauth2/token` requests through the actual handler (same harness as `oauth.test.ts`'s reuse-grace suite) with a fake `AUTH_EVENTS` binding, confirming each of the four events fires in the right scenario.
- `pnpm vitest run --config vitest.projects.ts --project @uploads/auth` — 30 files, 436 tests, all passing.
- `pnpm test` from repo root — 361 files, 5594 tests, all passing (no regressions).
- `pnpm exec tsc --noEmit` in apps/auth — clean.
- `wrangler types` regenerated `worker-configuration.d.ts` (gitignored) and confirms `AUTH_EVENTS: AnalyticsEngineDataset` is now in scope.

## Caveats for review

- The `ctx.context` snapshot-handoff technique (see above) is a documented reliance on better-auth 1.7.1's dispatcher internals — not a patch to the plugin, but worth a second look given how load-bearing it is for the replay/teardown split.
- `refresh_rotation_replay` fires whether the plugin actually served a cached replay response or — rarely, if its stored `rotationReplayResponse` went missing — failed anyway with `invalid_grant`. Both cases mean "a within-grace reuse was attempted," which is the #912 signal; I didn't split them into two events to keep the schema small, called out in the code comment.
- `client_id` is read from the request body only; a client authenticating via HTTP Basic auth instead of a body field won't populate it on some events (falls back to the D1 row's `clientId` when a snapshot exists, which covers the refresh-token cases).

### DIFF
--- a/apps/auth/src/auth.ts
+++ b/apps/auth/src/auth.ts
@@ -41,6 +41,11 @@ import { sendAuthEmail } from "./email";
 import { localDemoEnabled, localDemoPlugin } from "./local-demo";
 import { memberCapDenial } from "./member-cap";
 import { notifyAdminsOfMemberJoin } from "./notify-member-join";
+import {
+  refreshTokenAfterHook,
+  snapshotBeforeOAuthTokenRequest,
+  type AuthEventsEnv,
+} from "./oauth-observability";
 import { createDurableRateLimitStorage, type RateLimitNamespaceLike } from "./rate-limit";
 import * as schema from "./schema";
 import { stripePluginOrNone } from "./stripe-plugin";
@@ -310,11 +315,19 @@ async function applyOAuthClientInterop(
 function authBeforeHook(
   db: ReturnType<typeof drizzle<typeof schema>>,
   registeredScopesForClientId: (clientId: string | undefined) => Promise<readonly string[]>,
+  env: { DB: D1Database },
 ) {
   return createAuthMiddleware(async (ctx) => {
     const oauthOverride = await applyOAuthClientInterop(ctx, registeredScopesForClientId);
     if (oauthOverride) return oauthOverride;
     await enforceLastAdminGuard(ctx, db);
+    // Issue #912: snapshots a presented refresh token's D1 state ahead of
+    // oauthProvider()'s own /oauth2/token handler — see
+    // oauth-observability.ts's module doc comment for why this rides here
+    // instead of a dedicated hook slot (Better Auth's `hooks.before` takes
+    // exactly one handler). No-op for every path other than a
+    // grant_type=refresh_token request at /oauth2/token.
+    await snapshotBeforeOAuthTokenRequest(ctx, env);
   });
 }
 
@@ -409,7 +422,8 @@ async function enforceLastAdminGuard(
 }
 
 export type AuthEnv = GitHubCredentialsEnv &
-  DashApiKeyEnv & {
+  DashApiKeyEnv &
+  AuthEventsEnv & {
     DB: D1Database;
     EMAIL?: import("./email").EmailBinding;
     BETTER_AUTH_URL?: string;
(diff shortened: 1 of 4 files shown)


zachdunn:
Superseded by #914 (merged) — same design implemented from the same brief by a parallel agent. Closing as duplicate.

Issue #936, opened 2026-09-02
SVG/XML active content: follow-ups from review


zachdunn:
Follow-ups deferred from the #929 implementation's reviews. None block the feature; each is small and independent.

- **Stale host record on KV write failure.** `probeHostActiveContent` writes the record last; if `REGISTRY.put` itself throws after a failing probe, a previous day's `ok: true` record survives until the next sweep (≤ 24 h). Documented and tested; fix is to write a failing record first or retry the put once.
- **Reputation filter false positives.** `containsActiveMarkup`'s `\bon[a-z]+\s*=` also matches ordinary XML attributes (`online=`, `once=`), rejected with `reason: "active_markup"`. The CSP is the control, so the regex can be narrowed to a known event-attribute list (or `on[a-z]+` only inside a start tag) without weakening anything.
- **Entity-encoded evasion of the filter.** `&#106;avascript:` and `<set attributeName="onclick">` pass the filter. Same reasoning: the filter is reputation defense; decide whether to decode entities before scanning or leave as documented.
- **SVG image dimensions.** `detectImageDimensions` has no SVG branch, so gated SVGs get no `image.width`/`image.height` and the managed comment cannot size them. Parse `width`/`height`/`viewBox` from the root element.
- **Settings page tests.** The storage settings page's new "SVG & XML" row and Check-now button, and `verifyLaneActiveContent` in `api-client.ts`, have no automated coverage (the inline Astro script has no harness).
- **Ops:** the Transform Rule expression in `docs/ops.md` must be validated in the dashboard; if regex `matches` is unavailable on the plan, use the `eq` fallback documented there.

Refs: #929, the implementing PR.

zachdunn:
Two more from the final re-review:
- The settings page's SVG & XML row reads only the storage host's record; the gate also requires the embed twin. The page can say "Verified" while uploads 415 (fails closed; admin-only surface). Show both hosts or use the gate's result.
- The gate checks the workspace's host plus its embed twin but not the other stable twin (`store.uploads.sh` vs `storage.uploads.sh`). All three share one Transform Rule and one sweep today; requiring every hosted host to be fresh would close it outright.
- `admin-active-content-probe.test.ts` never pins that an `operator:read` token gets 403 on the POST (shared `adminAuth` covers it).

zachdunn:
Shipped:
- KV write retry for the host probe record: #941
[...]

PR #937, opened 2026-09-02
feat: accept SVG and XML on lanes verified to serve them sandboxed

### LINKED ISSUE #929: Accept SVG and XML behind a sandboxing CSP on the storage hosts
## Why

#928 opens uploads to PDF, zip, gzip, MOV, and text, but keeps SVG and XML out. SVG is the one media type GitHub's own attachment service accepts that we do not, and agents produce it (Mermaid and chart exports, architecture diagrams). XML shows up as JUnit and coverage reports. Both should be a lane we serve, done responsibly.

## The problem

Served inline, an SVG or XML document can run script: `<script>`, `on*` handlers, `<foreignObject>` with XHTML, `<?xml-stylesheet?>` XSLT, `javascript:` hrefs. Today the stored content type is the only control this repo owns on `storage.uploads.sh` / `embed.uploads.sh` (bare R2 custom domains, no Worker), which is why `guards.ts` rejects `image/svg+xml`. Embedding via `` never runs script; the risk is a person navigating to the URL directly, on an origin we own.

Sanitizing is not the answer: Workers have no DOM, allowlist rewriters get bypassed, and a sanitizer would have to be right forever.

## Proposal: the raw.githubusercontent.com model

Serve SVG and XML with a sandboxing CSP at the edge so the document renders in an opaque origin.

1. **Zone Transform Rule (ops, same mechanism as the `embed.uploads.sh` Cache-Control rule in docs/ops.md).** On `storage.uploads.sh`, `embed.uploads.sh`, and `store.uploads.sh`, for responses whose `Content-Type` starts with `image/svg+xml`, `application/xml`, or `text/xml`, set:
   - `Content-Security-Policy: default-src 'none'; style-src 'unsafe-inline'; img-src data:; sandbox`
   - `X-Content-Type-Options: nosniff`
   With `sandbox`, script cannot run and the document has an opaque origin, so nothing on our hosts is reachable even if the SVG is hostile. This is what GitHub does for raw user content.
2. **Verify before trusting.** A small `_internal/probe.svg` object per host and a check (post-deploy step or scheduled) that fetches it and asserts both headers. The allowlist gains `image/svg+xml`, `application/xml`, and `text/xml` only behind a flag (`FLAGS` binding, like the poster gate) that ops flips after the probe is green. If the probe ever fails, the flag goes off and new SVG uploads 415 again; already-stored objects keep serving but with the same edge rule they always had.
3. **Hosted lanes only.** BYO buckets (S3 or a customer R2 on their own domain) cannot get the header from us. SVG/XML stay rejected on BYO lanes. A later opt-in could let a workspace owner attest they set the header on their domain; not in scope.
4. **Guard changes** (`apps/api/src/guards.ts`): sniff SVG as `<svg` or `<?xml` … `<svg` within the first 1 KiB after BOM, whitespace, and comments; XML is declared-plus-plausible like the text types (`looksLikeText` plus the body starts with `<`). `contentTypeFromKey` gains `svg`, `xml`. Keep a cheap reputation pre-filter that rejects bodies containing `<script`, `on[a-z]+=`, `javascript:`, or `<foreignObject` and say plainly in the comment that the CSP is the control, not this filter.
5. **Web:** `fileKind("image/svg+xml")` becomes `image` (retire the `unsupported` branch in `MediaStage`); `thumbUrl` keeps skipping `.svg` (cdn-cgi does not transform it); the `/f/` page renders SVG through ``, never inline.
6. **Comment renderer:** SVG already maps to `image/*` in `inferContentType`, so it embeds as ``; GitHub's Camo proxies it with its own CSP. Confirm Camo serves SVG from our hosts before the copy says so.
7. **Copy:** limits page, `put` help, both skills; note that inline viewing is sandboxed (no scripts, no external resources) so interactive SVGs will not be interactive.

## Acceptance

- [ ] Transform Rule live on all three hosts; probe asserts CSP + nosniff
- [ ] Flag-gated allowlist entries for `image/svg+xml`, `application/xml`, `text/xml`, hosted lanes only
- [ ] Guard tests: sniffed SVG accepted when flag on, 415 when off; XML declared path; pre-filter rejects a scripted SVG; BYO lane rejects regardless of flag
- [ ] `/f/` renders SVG via ``; navigating to the storage URL of a scripted SVG shows no alert in Chrome, Firefox, Safari
- [ ] Copy updated

## Out of scope

- Sanitizer-based acceptance without the CSP
- BYO-lane SVG
- HTML (same mechanism would work, but there is no demand and the phishing surface is different)

Related: #925, #928, #410.

zachdunn:
Closes #929. Design: [spec](docs/superpowers/specs/2026-09-02-svg-xml-active-content-design.md) · [plan](docs/superpowers/plans/2026-09-02-svg-xml-active-content.md).

SVG, `application/xml`, and `text/xml` are accepted on a storage lane only while that lane's public host is verified to serve SVG with a sandboxing CSP and `nosniff`. The header is the control (the raw.githubusercontent.com model); nothing in this PR sanitizes.

## How it works

- **One probe, every lane.** `probeActiveContent` writes an inert SVG and an inert XML object under `_internal/uploads-verify/`, fetches each through the lane's public base URL, and accepts only when both echo their content type, carry a `sandbox` CSP directive (no `allow-scripts` / `allow-same-origin`, one CSP header), and `X-Content-Type-Options: nosniff`. A thrown fetch or write is `inconclusive` and never enables. Orphaned probe objects are reaped daily.
- **Hosted lanes:** ops sets the header with a zone Transform Rule (`docs/ops.md`). A daily cron probes `storage.uploads.sh`, `store.uploads.sh`, and `embed.uploads.sh` and writes KV host records; `POST /admin/active-content/probe` (operator token) runs it on demand. The gate requires fresh (≤ 48 h) `ok` records for **all three** hosted hosts, since they front the same bucket.
- **BYO lanes:** the owner sets the header any way they like. The probe runs inside lane verify (save and activate) and from a "Check now" button on the storage settings page; the stamp is good for 30 days. Any sandboxing CSP passes.
- **Gate:** `activeContentAllowed(env, ws)` = workspace opt-out → Flagship flag `active-content-uploads` (fail-closed) → lane health → lane/host freshness. `resolveUploadPolicy(record, { activeContent })` makes the second argument required and strips the gated rows from every allowlist, including an `allowedContentTypes` override.
- **Guard:** the three types are declared-only, never sniffed, with per-row plausibility (SVG root must be `<svg`; XML must start with `<`), a 4 MiB row cap checked before the body scan, and a reputation pre-filter (`<script`, `on*=`, `javascript:`, `<foreignObject`, XSLT) that returns `reason: "active_markup"`. Unverified lanes return `reason: "lane_not_verified"`. Server-side copies (attach, promote, prefix rotation) carry `serverCopy` and skip only the freshness checks (same lane, same host, no new exposure); the kill switch, workspace opt-out, and lane health still apply, and promote/rotation skip-and-record a rejected object rather than wedging. The gate is evaluated lazily, only when the declared type is gated, and once per MCP batch.
- **Web:** `/f/` renders SVG through `` only; `thumbUrl` still skips `.svg`; settings shows per-lane readiness from the same status the gate uses (with a reason when closed) and a Check-now button with a 60 s per-lane cooldown. Downloads via `?download=1` carry `nosniff` and a `sandbox` CSP; presigned URLs for gated types expire in 15 minutes. Docs: BYO setup step 5, limits page, ops runbook.

## Rollout (flag stays off at merge)

1. Merge. With the flag off, SVG/XML keep 415ing exactly as today (now with `reason: "lane_not_verified"`). The only always-on behavior is three probe objects written and deleted daily in the shared bucket.
2. Add the Transform Rule to all three hosted hosts; validate the expression in the rule editor (`docs/ops.md` has the escaped regex and an `eq` fallback), then `curl -I` a real `.svg` on each host.
3. `curl -X POST -H "Authorization: Bearer $ADMIN_TOKEN" https://api.uploads.sh/admin/active-content/probe` and confirm three `ok: true` records.
4. Turn on `active-content-uploads`. Upload an inert SVG and open it in Chrome, Firefox, and Safari: `window.origin === "null"`, no script. Upload one with `<script>` and expect 415 `active_markup`. Confirm Camo renders an SVG in a PR comment.
5. The flag is the rollback.

## Review trail

Four batches, each reviewed (Opus on the guard batches), one fix round on the guard batch, a whole-branch Opus review with one fix wave, a four-angle simplify pass with one fix round, and an adversarial security review (Opus; Codex refused the brief under its cyber policy) with one fix wave. Every fix wave was re-reviewed. Deferred minors are collected in #936.

### DIFF
--- a/.changeset/svg-xml-active-content.md
+++ b/.changeset/svg-xml-active-content.md
@@ -1,0 +1,5 @@
+---
+"@buildinternet/uploads": minor
+---
+
+`put` and `attach` accept SVG and XML (`image/svg+xml`, `application/xml`, `text/xml`) on a storage lane once its public host is verified to serve them behind a sandboxing Content-Security-Policy. An unverified lane keeps 415ing these types. The managed comment embeds SVG as `<img>`. Requires the matching platform release.
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -12,6 +12,7 @@
     "./file-metadata": "./src/file-metadata.ts",
     "./file-search": "./src/file-search.ts",
     "./guards": "./src/guards.ts",
+    "./active-content": "./src/active-content.ts",
     "./budget": "./src/budget.ts",
     "./usage": "./src/usage.ts",
     "./reconcile": "./src/reconcile.ts",
--- a/apps/api/src/active-content-hosts.test.ts
+++ b/apps/api/src/active-content-hosts.test.ts
@@ -1,0 +1,221 @@
+import { describe, expect, it, vi } from "vitest";
+import {
+  HOSTED_ACTIVE_CONTENT_HOSTS,
+  hostActiveContentKey,
+  probeHostActiveContent,
+  readHostActiveContent,
+  runActiveContentHostSweep,
+} from "./active-content-hosts";
+import { FakeR2Bucket } from "../test/fake-r2";
+
+/**
+ * `vi.fn`'s inferred call signature never structurally matches the
+ * overloaded `typeof fetch`, so every fake is built with an explicit,
+ * fetch-compatible signature and handed to the functions under test through
+ * this cast rather than fighting TS overload assignability at each call
+ * site — the fake's own `.mock.calls` stay precisely typed for assertions.
+ */
+function asFetch(
+  fn: ReturnType<typeof vi.fn<(url: string, init?: RequestInit) => Promise<Response>>>,
+): typeof fetch {
+  return fn as unknown as typeof fetch;
+}
+
+/** In-memory REGISTRY KV stand-in — get/put/delete over a Map. */
+function fakeRegistry(records: Record<string, unknown> = {}) {
+  const store = new Map(Object.entries(records));
+  return {
+    store,
+    get: (async (key: string) => store.get(key) ?? null) as unknown as KVNamespace["get"],
+    put: vi.fn(async (key: string, value: string) => {
+      store.set(key, JSON.parse(value));
+    }) as unknown as KVNamespace["put"],
+    delete: (async (key: string) => {
+      store.delete(key);
+    }) as unknown as KVNamespace["delete"],
+  };
+}
+
+const goodHeaders = {
+  "content-type": "image/svg+xml",
+  "content-security-policy": "default-src 'none'; sandbox",
+  "x-content-type-options": "nosniff",
+};
+
+function env(over: Record<string, unknown> = {}) {
+  return {
+    UPLOADS_DEFAULT: new FakeR2Bucket(),
+    REGISTRY: fakeRegistry(),
+    ...over,
+  } as unknown as Env;
+}
+
+describe("hostActiveContentKey", () => {
+  it("namespaces the KV key by host", () => {
+    expect(hostActiveContentKey("storage.uploads.sh")).toBe(
+      "host-active-content:storage.uploads.sh",
+    );
+  });
+});
+
+describe("HOSTED_ACTIVE_CONTENT_HOSTS", () => {
+  it("is the deduped hosted host set", () => {
+    expect([...new Set(HOSTED_ACTIVE_CONTENT_HOSTS)]).toEqual([...HOSTED_ACTIVE_CONTENT_HOSTS]);
+    expect(HOSTED_ACTIVE_CONTENT_HOSTS).toEqual(
+      expect.arrayContaining(["storage.uploads.sh", "store.uploads.sh", "embed.uploads.sh"]),
+    );
+    expect(HOSTED_ACTIVE_CONTENT_HOSTS.length).toBe(3);
+  });
+});
+
+describe("probeHostActiveContent", () => {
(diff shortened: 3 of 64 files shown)


zachdunn:
@coderabbitai review

zachdunn:
Rollout completed 2026-09-02:

- Transform Rules applied (nosniff everywhere on the storage hosts; sandbox CSP on SVG/XML). Operator probe: `ok: true` on `storage.`, `store.`, and `embed.uploads.sh`.
- Flag `active-content-uploads` on.
- Inert SVG → 201 `image/svg+xml`; JUnit XML → 201 `application/xml`; SVG with `<script>` → 415.
- All three hosts serve both types with `Content-Security-Policy: default-src 'none'; style-src 'unsafe-inline'; img-src data:; sandbox` and `X-Content-Type-Options: nosniff`.
- Chromium, navigated to the storage URL directly: `window.origin === "null"`, `document.cookie` throws SecurityError, SVG renders, no script elements. Firefox and Safari were not exercised from this environment.
- `/f/` page renders the SVG via `` only (no object/embed/iframe).

PR #942, opened 2026-09-02
fix(api): narrow the active-markup reputation filter for gated SVG/XML

zachdunn:
Narrows `containsActiveMarkup` in `apps/api/src/guards.ts`, the reputation pre-filter for gated SVG/XML uploads. The actual control is the sandboxing CSP the serving lane sends; this filter only needs to reject obvious active content without false positives.

- Event-handler attribute matching now uses a fixed list of known `on*` handler names instead of the broad `on[a-z]+` pattern, so ordinary attributes like `online=`/`once=` no longer trip a 415.
- Numeric and named (`&lt; &gt; &amp; &quot; &apos;`) XML entities are decoded before scanning, so `&#106;avascript:` is caught.
- SMIL `<set>`/`<animate>` tags that retarget an event attribute via `attributeName="onclick"` are now rejected.
- Doc comment above the function now states what it does and does not promise.
- Existing 415 `details.reason: "active_markup"` behavior is untouched; no route files were changed.

Refs #936

Test command: `pnpm --filter @uploads/api test` — 183 files, 2772 tests passed.

### DIFF
--- a/.changeset/svg-image-dimensions.md
+++ b/.changeset/svg-image-dimensions.md
@@ -1,0 +1,5 @@
+---
+"@buildinternet/uploads": patch
+---
+
+Read width/height for SVG uploads from the root `<svg>` tag's `width`/`height` attributes (or `viewBox` as a fallback), so `image.width`/`image.height` server metadata — and the managed GitHub comment's sizing — now cover SVGs the same as other image types.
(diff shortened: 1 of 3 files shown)


zachdunn:
Note for history: the squash commit for this PR (cada6181) carries the message "detect SVG dimensions" but its diff is the `containsActiveMarkup` narrowing (event-attribute allowlist, entity decoding, SMIL `<set>` coverage). The two #936 branches were cross-wired at push time; the SVG dimensions code lands in #943. Both changeset entries describe the right change, so the changelog will read correctly.

PR #943, opened 2026-09-02
feat(api): detect SVG dimensions for image metadata

zachdunn:
## Summary

Adds an `image/svg+xml` branch to `detectImageDimensions` (`apps/api/src/guards.ts`) so SVG uploads get `image.width`/`image.height` server metadata like the other image types, letting the managed GitHub comment size them.

- Decodes at most the first 8 KiB as UTF-8 and finds the root `<svg …>` start tag.
- Reads `width`/`height` when both are plain numbers or have a `px` suffix.
- Falls back to `viewBox`'s third/fourth values (rounded) when `width`/`height` aren't usable.
- Percent, em, or other units, or no parseable `<svg` tag → `undefined` (fail-open, matching the existing contract).

Callers verified:
- `apps/api/src/files-core.ts`'s `uploadKind(contentType) !== "image"` gate already passes SVG through — `image/svg+xml`'s row in `GATED_UPLOAD_TYPES` has `kind: "image"`.
- `apps/api/src/github-ingest.ts` never reaches `detectImageDimensions` for SVG today: it calls with `activeContent: false`, and `detectContentType` (magic-byte sniff) returns `null` for SVG since it's a declared-only type, so the `!sniffed` check skips it before the allowlist/dimension step. No change needed there; noting it as a known gap outside this issue's scope.
- `packages/comment-render`: `naturalMediaWidth`/`isSmallAsset` already work generically off `dims.width`/`dims.height` — no changes needed.

## Tests

Added cases to `apps/api/test/guards.test.ts`'s `detectImageDimensions` describe block: numeric width/height, `px`-suffixed width/height, `viewBox` fallback, percent units → `undefined`, no `<svg` root → `undefined`.

```
pnpm --filter @uploads/api test
```
2773 tests passed (183 files).

Refs #936

### DIFF
--- a/.changeset/active-markup-filter-precision.md
+++ b/.changeset/active-markup-filter-precision.md
@@ -1,0 +1,5 @@
+---
+"@buildinternet/uploads": patch
+---
+
+Sharpen the server-side reputation pre-filter for gated SVG/XML uploads (`containsActiveMarkup`): event-handler attribute matching no longer false-positives on ordinary attributes like `online=`/`once=`, and entity-encoded evasions (`&#106;avascript:`, SMIL `<set attributeName="onclick">`) are now rejected. The actual security control remains the sandboxing CSP on the serving lane; this filter is defense in depth.
--- a/apps/api/src/guards.ts
+++ b/apps/api/src/guards.ts
@@ -525,12 +525,16 @@ export interface ImageDimensions {
   height: number;
 }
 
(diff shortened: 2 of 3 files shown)


zachdunn:
Note for history: squash commit e3b1f85b says "narrow the active-markup reputation filter" but its diff is the SVG dimensions change (the branch commit message was used for the squash, not the PR title). The filter change is cada6181 (#942).

PR #1000, opened 2026-09-17
feat(mcp): fetch AS JWKS over the AUTH service binding for OAuth token verification

zachdunn:
Follow-up to #999 — the second (and last) first-party service-to-service call that was going over HTTP instead of a service binding.

## What

The MCP worker verifies OAuth 2.1 access tokens by fetching the authorization server's JWKS. It did this over plain `fetch` to `${AUTH_ORIGIN}/api/auth/jwks` (default `https://uploads.sh/api/auth/jwks`) — even though the worker already has an `AUTH` service binding to `uploads-auth` it wasn't using here.

Route that JWKS fetch over the binding.

## Why

- **Resilience** — token verification no longer depends on the public edge. The exact failure mode from the #997 incident (a WAF/bot rule, or any edge hiccup) would previously have broken MCP OAuth auth; over the binding it can't.
- **One fewer hop** — the default issuer origin is `uploads.sh`, so today the JWKS fetch is web `/api/auth` proxy → auth worker. `env.AUTH.fetch("https://auth.internal/api/auth/jwks")` goes straight to `uploads-auth`.
- Cheaper (no billable edge subrequest), though low-volume given the 5-min in-isolate JWKS cache.

## How

`jwksFetcherFor(env)` in `apps/mcp/src/oauth.ts` returns the `JwksFetcher` to use: prefers `env.AUTH` (rewriting the JWKS URL's host to the `https://auth.internal` alias while preserving the `/api/auth/jwks` path — the binding routes by path, matching apps/api's `session-auth.ts` convention) and falls back to plain `fetch` against the public URL when the binding is absent (local dev). `oauthAuth` threads it in via the existing `jwksFetcher` config seam. Mirrors the binding-else-HTTP pattern in apps/web's `api-proxy.ts` and #999.

The `iss` claim is unchanged — only the JWKS transport moves; token issuer validation still matches the public issuer.

## Notes

- Refreshes the module docblock (previously "The MCP worker has no service binding to apps/auth", now stale — the binding has existed in `wrangler.jsonc`).
- The OAuth test suite previously stubbed JWKS via a global-`fetch` override and noted the `jwksFetcher` seam was unthreaded; it now threads through the AUTH binding, so the fake `env.AUTH` serves the key set at `/api/auth/jwks` (shared `authStub` helper). New `apps/mcp/test/oauth.test.ts` unit-covers `jwksFetcherFor` (binding routing, dev fallback, error handling).

## Audit result

This completes the sweep prompted after #999: **api ↔ auth already use their bindings**, web's proxies + #999 cover web, and `apps/auth/src/cimd-transport.ts` fetches *third-party* client metadata (correctly HTTP, must not use a binding). MCP's JWKS was the one remaining first-party HTTP call — this is it.

## Verification

- `pnpm --filter @uploads/mcp test` → 133 passed
- `pnpm --filter @uploads/mcp typecheck` → pass
- Binding path best exercised by this PR's Workers Preview against production `uploads-auth`.


zachdunn:
@coderabbitai review

